PR #20, opened 2015-01-03
Features/talk version history

mattstauffer:


mattstauffer:
This issue was moved to tightenco/symposium#20

PR #53, opened 2015-05-08
Add rudimentary bio management; Fixes #1

mattstauffer:
image

image

image

PR #54, opened 2015-05-08
Add gravatar

mattstauffer:
Builds on #53


mattstauffer:
Crap. Broke this with git mismanagement. Need to re-write and re-pull. Argh.

PR #55, opened 2015-05-10
Bios and gravatar

### LINKED ISSUE #1: Add bios
Different sizes, and versions, of bios are necessary for conference submissions

mattstauffer:
#54 and #53 merged after drama w git.

Closes #1 

- Add bios
- Add gravatar in header

image

screen shot 2015-05-10 at 6 57 25 pm

screen shot 2015-05-10 at 6 57 49 pm

screen shot 2015-05-10 at 6 58 24 pm


app/Http/Controllers/BiosController.php
```diff
@@ -0,0 +1,144 @@
...
+{
+    // @todo: Why is this here but only used for update? Why are they not using the same pattern?
+    protected $validation = [
+    ];
+
+    public function index()
+    {
+        $bios = Bio::where('user_id', Auth::user()->id)
+            ->orderBy('nickname')
+            ->get();
```
adamwathan: Should use a `bios` relation on the `User` instead of looking up bios based on `user_id`. Leaky abstraction, etc.

app/Http/routes.php
```diff
@@ -52,7 +52,9 @@
     // Necessary because IDK and please let's fix this @todo
     Route::post('talks/{id}', 'TalksController@update');
     Route::post('conferences/{id}', 'ConferencesController@update');
+    Route::post('bios/{id}', 'BiosController@update');
```
adamwathan: The `update` action should be a `PUT` and should be handled by the resource declaration no?

app/Http/Controllers/BiosController.php
```diff
@@ -0,0 +1,144 @@
...
+            $bio = $form->complete();
+        } catch (ValidationException $e) {
+            return Redirect::to('bios/create')
+                ->withErrors($e->errors())
+                ->withInput();
+        }
+
+        Session::flash('message', 'Successfully created new bio.');
+
+        return Redirect::to('/bios/' . $bio->id);
```
adamwathan: We should use named routes for this stuff, the `resource` routing style still gives you route names. This one would be `bios.show`.

app/Http/Controllers/BiosController.php
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    public function show($id)
+    {
+        try {
+            $bio = Bio::where('id', $id)->firstOrFail();
+        } catch (Illuminate\Database\Eloquent\ModelNotFoundException $e) {
```
adamwathan: Should import this class.

app/Http/Controllers/BiosController.php
```diff
@@ -0,0 +1,144 @@
...
+
+        Session::flash('message', 'Successfully created new bio.');
+
+        return Redirect::to('/bios/' . $bio->id);
+    }
+
+    public function show($id)
+    {
+        try {
+            $bio = Bio::where('id', $id)->firstOrFail();
```
adamwathan: This is equivalent to `Bio::findOrFail($id)`;

app/Http/Controllers/BiosController.php
```diff
@@ -0,0 +1,144 @@
...
+        } catch (Illuminate\Database\Eloquent\ModelNotFoundException $e) {
+            Session::flash('error-message', 'Sorry, but that isn\'t a valid URL.');
```
adamwathan: I like double quotes for human-meant strings so you can avoid escaping single quotes (since they are really common in human-readable strings). Also do interpolation often in human-readable string so that's nice. Just a thought but I find it to be a useful convention.

Computer-meant strings = single quotes (array keys, route names, controller action bindings, etc.)
Human-meant strings = double quotes (flash messages, human-readable log messages, etc.)

app/Http/Controllers/BiosController.php
```diff
@@ -0,0 +1,144 @@
...
+        }
+
+        return View::make('bios.show')
+            ->with('bio', $bio);
+    }
+
+    public function edit($id)
+    {
+        try {
+            $bio = Bio::where('id', $id)->firstOrFail();
```
adamwathan: findOrFail

app/Http/Controllers/BiosController.php
```diff
@@ -0,0 +1,144 @@
...
+        } catch (Illuminate\Database\Eloquent\ModelNotFoundException $e) {
```
adamwathan: Import namespace

app/Http/Controllers/BiosController.php
```diff
@@ -0,0 +1,144 @@
...
+            Session::flash('error-message', 'Sorry, but that isn\'t a valid URL.');
+            Log::error($e);
+            return Redirect::to('/');
+        }
+
+        if ($bio->user_id != Auth::user()->id) {
+            Session::flash('error-message', 'Sorry, but you don\'t own that bio.');
+            Log::error('User ' . Auth::user()->id . ' tried to edit a bio they don\'t own.');
+            return Redirect::to('/');
+        }
```
adamwathan: This whole thing can be skipped by using the strategy I mentioned in the earlier comment. I'd just 404 here, like when you try to visit a private GitHub repo you don't own.

app/Http/Controllers/BiosController.php
```diff
@@ -0,0 +1,144 @@
...
+        return View::make('bios.edit')
+            ->with('bio', $bio);
+    }
+
+    public function update($id)
+    {
+        $data = Input::all();
+
+        // Make validator
+        $validator = Validator::make($data, $this->validation);
```
adamwathan:
```
// Adds 5 and 4 and puts them in $x
$x = 5 + 4;
```

app/Http/Controllers/BiosController.php
```diff
@@ -0,0 +1,144 @@
...
+        if ($validator->passes()) {
+            // Pull
+            $bio = Bio::where('id', $id)->firstOrFail();
+
+            // Validate ownership
+            if ($bio->user_id != Auth::user()->id) {
+                Session::flash('error-message', 'Sorry, but you don\'t own that bio.');
+                Log::error('User ' . Auth::user()->id . ' tried to edit a bio they don\'t own.');
+                return Redirect::to('/');
+            }
```
adamwathan: Can do another `$user->bios()->findOrFail($id)` here.

app/Http/Controllers/BiosController.php
```diff
@@ -0,0 +1,144 @@
...
+
+            // Save
+            $bio->body = Input::get('body');
+            $bio->user_id = Auth::user()->id;
```
adamwathan: Don't need this line, the `$user_id` property will already be set. If it wasn't, `$user->bios()->save($bio)` would be better, and `$user->addBio($bio)` even better IMO.

app/Http/Controllers/BiosController.php
```diff
@@ -0,0 +1,144 @@
...
+        }
+
+        return Redirect::to('bios/' . $id . '/edit')
+            ->withErrors($validator)
+            ->withInput();
+    }
+
+    public function destroy($id)
+    {
+        $bio = Bio::where('id', $id)->firstOrFail();
```
adamwathan: `$user->bios->findOrFail($id)` :+1:

app/Http/Controllers/BiosController.php
```diff
@@ -0,0 +1,144 @@
...
+
+        // Validate ownership
+        if ($bio->user_id != Auth::user()->id) {
+            Session::flash('error-message', 'Sorry, but you don\'t own that bio.');
+            Log::error('User ' . Auth::user()->id . ' tried to delete a bio they don\'t own.');
+            return Redirect::to('/');
+        }
+
+        Session::flash('success-message', 'Bio successfully deleted.');
+        $bio->delete();
```
adamwathan: Maybe flash after instead of before? No _great_ reason, could argue you would want to check the return of `delete` to make sure it deleted, or avoid flashing the message if there was some bizarre database failure that threw an exception, but really it just bothers me in an OCD sort of way to say that it was deleted before you delete it, haha...

app/modelEvents.php
```diff
@@ -1,7 +1,7 @@
 <?php
 
 // UUID-based models
-$models = ['Talk', 'Conference', 'TalkVersion', 'TalkVersionRevision'];
+$models = ['Talk', 'Conference', 'TalkVersion', 'TalkVersionRevision', 'Bio'];
 
 foreach ($models as $model) {
     $model::creating(function ($model) {
```
adamwathan: Not new to this PR, but I think it makes more sense to put the UUID generation in a `save` override on the `UuidBase` model like we ended up doing on other projects, especially since we have that base model anyways. This separate-file-model-event-binding stuff was really just a workaround for weird PHPUnit static state crap anyways.

app/models/Bio.php
```diff
@@ -0,0 +1,24 @@
...
+use Carbon\Carbon;
+use Illuminate\Database\Eloquent\Collection;
+
+class Bio extends UuidBase
+{
+    protected $table = 'bios';
+
+    protected $guarded = array(
+        'id'
+    );
```
adamwathan: Better to _just_ use `$fillable` I think. I dunno what the precedence rules are between the two and trying to keep the sum of them as a complete representation of the database schema is a losing battle. Also 5.3 arrays BURN IN HELL.

app/models/Bio.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+use Carbon\Carbon;
+use Illuminate\Database\Eloquent\Collection;
+
+class Bio extends UuidBase
```
adamwathan: Non-namespaced model makes me cry a little tear of joy. The good ol' days.

composer.json
```diff
@@ -7,7 +7,11 @@
         {
             "type": "git",
             "url": "git@github.com:mattstauffer/joindin-client.git"
-        }
+        },
+		{
+			"type": "git",
+			"url": "git@github.com:mattstauffer/laravel-gravatar.git"
+		}
```
adamwathan: Tabs COME ON.

composer.json
```diff
@@ -17,7 +21,8 @@
         "rexxars/joindin-client": "dev-master",
         "guzzlehttp/guzzle": "~5.0@dev",
         "illuminate/html": "~5.0",
-        "doctrine/dbal": "~2.6@dev"
+        "doctrine/dbal": "~2.6@dev",
+		"thomaswelton/laravel-gravatar": "dev-laravel5-support"
```
adamwathan: TABS!!!

config/app.php
```diff
@@ -195,6 +196,7 @@
 		// 'Bugsnag'   => 'Bugsnag\BugsnagLaravel\BugsnagFacade',
 		'Form'      => 'Illuminate\Html\FormFacade',
 		'HTML'      => 'Illuminate\Html\HtmlFacade',
+	    'Gravatar'  => 'Thomaswelton\LaravelGravatar\Facades\Gravatar'
```
adamwathan: More mixed spaces and tabs...

public/js/main.js
```diff
@@ -59,4 +59,9 @@ $(function() {
 			}
 		});
 	});
+
+	$('.bio-modal').on('shown.bs.modal', function () {
+  		$(this).find('textarea').focus();
+  		console.log('t');
```
adamwathan: This is a handy little gulp plugin: https://www.npmjs.com/package/gulp-strip-debug

resources/sass/app.scss
```diff
@@ -242,6 +250,14 @@ legend {
     }
 }
 
+.list-bios {
+    @extend %index-list;
```
adamwathan: I would say should just use the same component class for both of these if one is extending the other, and add another class for the `h3`s.

Makes me think of the ever brilliant Nicholas Gallagher article: http://nicolasgallagher.com/about-html-semantics-front-end-architecture/

> The most reusable components are those with class names that are independent of the content.

resources/views/bios/create.blade.php
```diff
@@ -0,0 +1,26 @@
...
+            <div class="col-md-6">
+                <h1>Create Bio</h1>
+
+                <ul class="errors">
+                    @foreach ($errors->all() as $message)
+                        <li>{{ $message }}</li>
+                    @endforeach
+                </ul>
+
+                {{ Form::open(array('action' => 'BiosController@store', 'class' => 'new-bio-form')) }}
```
adamwathan: 5.3 array can be 5.4 array OR EVEN 5.6 ARRAY IF THERE WAS SUCH A BEAUTIFUL THING.

Greenfield project bleeding edge PHP versions FTW.

resources/views/bios/create.blade.php
```diff
@@ -0,0 +1,26 @@
...
+
+                @include('partials.bioform')
+
+                {{ Form::submit('Create', ['class' => 'btn btn-primary']) }}<br><br>
```
adamwathan: Still watching the world burn with those line breaks I see.

resources/views/bios/edit.blade.php
```diff
@@ -0,0 +1,26 @@
...
+            <div class="col-md-6">
+                <h1>Edit Bio</h1>
+
+                <ul class="errors">
+                    @foreach ($errors->all() as $message)
+                        <li>{{ $message }}</li>
+                    @endforeach
+                </ul>
+
+                {{ Form::open(array('action' => array('BiosController@update', $bio->id), 'class' => 'edit-bio-form')) }}
```
adamwathan: 5.3 array

resources/views/bios/edit.blade.php
```diff
@@ -0,0 +1,26 @@
...
+
+                @include('partials.bioform')
+
+                {{ Form::submit('Update', ['class' => 'btn btn-primary']) }}<br><br>
```
adamwathan: <br><br> like the sound I make when seeing line breaks used for styling chills my soul to it's frigid core.

resources/views/bios/index.blade.php
```diff
@@ -0,0 +1,51 @@
...
+                      </div>
+                    </div>
+
+                    <p>
+                    <a href="/bios/{{ $bio->id }}/delete" data-confirm="confirm"><button type="button" class="btn btn-xs btn-danger">Delete</button></a>
+                    <a href="/bios/{{ $bio->id }}/edit"><button type="button" class="btn btn-xs btn-default">Edit</button></a>
+                    <a href="#" data-toggle="modal" data-target="#modal-{{ $bio->id }}"><button type="button" class="btn btn-xs btn-default">Copy</button></a>
+                    </p>
+                </li>
+            @endforeach
```
adamwathan: Could actually use a `forelse` the way you've laid this one out :facepunch:

resources/views/bios/index.blade.php
```diff
@@ -0,0 +1,51 @@
...
+
+        <ul class="list-bios">
+            @if ($bios->count() == 0)
+                <li>No bios yet.</li>
+            @endif
+            @foreach ($bios as $bio)
+                <li>
+                    <h3><a href="/bios/{{ $bio->id }}">{{ $bio->nickname }}</a></h3>
+                    <?php /* TODO: cleaner substr */ ?>
+                    <p>{{ substr($bio->body, 0, 100) }}...</p>
```
adamwathan: Could definitely add an accessor to the `Bio` class for now that's like `preview` or something. Maybe a presenter down the road if there's starts being lots of things like that.

resources/views/conferences/edit.blade.php
```diff
@@ -13,7 +13,7 @@
                     @endforeach
                 </ul>
 
-                {{ Form::open(array('action' => array('ConferencesController@update', $conference->id), 'class' => 'edit-talk-form')) }}
+                {{ Form::open(array('action' => array('ConferencesController@update', $conference->id), 'class' => 'edit-conference-form')) }}
```
adamwathan: 5.3 array

resources/views/partials/nav.blade.php
```diff
@@ -1,10 +1,18 @@
...
     <ul class="primary-header__meta-nav">
         @if (Auth::check())
+            <li><a href="/bios">Bios</a></li>
             <li><a href="/conferences">Conferences</a></li>
             <li><a href="/talks">Talks</a></li>
-            <li><a href="/account">Account</a></li>
-            <li><a href="/log-out">Log out</a></li>
+            <li class="dropdown" role="presentation">
+                <a class="dropdown-toggle" data-toggle="dropdown" href="#" role="button" aria-expanded="false">
+                    <img src="{{ Gravatar::src(\Auth::user()->email) }}" class="gravatar"> Me <span class="caret"></span>
```
adamwathan: Is the leading slash in front of `Auth` needed?

app/Http/routes.php
mattstauffer: Yah, that's why I was confused when I first wrote this. I didn't know why the Laravel HTML Form generator was submitting as POST and PUT and I didn't feel like fighting it. But yes, we need to fix/figure that out. I'll try to dig into it later.

app/Http/Controllers/BiosController.php
mattstauffer: :+1:

app/Http/Controllers/BiosController.php
mattstauffer: :+1:

app/Http/Controllers/BiosController.php
mattstauffer: Your face is equivalent to `Bio::findOrFail($id);`

app/Http/Controllers/BiosController.php
mattstauffer: I like that.

app/Http/Controllers/BiosController.php
mattstauffer: duh. nice.

app/Http/Controllers/BiosController.php
mattstauffer: shut up beavis

app/Http/Controllers/BiosController.php
mattstauffer: huh, yah, dunno how that even snuck in there

app/Http/Controllers/BiosController.php
mattstauffer: Haha, no, I definitely get that.

app/modelEvents.php
mattstauffer: Agreed!

app/models/Bio.php
mattstauffer: Oh yah. how the heck did that even get in there?

app/models/Bio.php
mattstauffer: bahaha

composer.json
mattstauffer: OMG HOW DID THIS HAPPEN IM EVEN USING SUBLIME TEXT

config/app.php
mattstauffer: bahaha

public/js/main.js
mattstauffer: yah, need to move this thing to elixir soon

resources/sass/app.scss
mattstauffer: well, it's not that one is extending another; rather, there's an index-list base class that the other lists basically extend. It's just a lazy way to avoid having to do `class="index-list bios-list"` and then later `index-list conferences-list`, etc.

`index-list` is the class that's independent of the content, and then it's just mildly extended in each context.

Thoughts?

resources/views/bios/create.blade.php
mattstauffer: hahaha :) greenfield project that's copied from a project I wrote in 2013... https://github.com/mattstauffer/markedstyle/commit/9b1d7117486d86f3e0eb88866138f9336420ae78

:D agreed though

resources/views/bios/index.blade.php
mattstauffer: BOOM.

i just couldn't remember what it was because I wrote this on a plane or something. Thanks for the reminder to go use it. :)

resources/views/bios/index.blade.php
mattstauffer: :+1:

resources/views/partials/nav.blade.php
mattstauffer: highly doubtful

mattstauffer:
@adamwathan all input fixed or moved here: https://github.com/tightenco/symposium/issues/57

app/Http/routes.php
adamwathan: https://github.com/laravel/framework/issues/616

Basically the form builder doesn't infer the method when you use the action or route options (unfortunately). So can resolve easily by just adding a `'method' => 'put'` as part of the array when opening the form on the edit pages.

resources/sass/app.scss
adamwathan: Yeah I think the multi-class pattern is the better approach generally. To me it's analogous to:

```
.btn {
 // main button styles
}

.btn--success {
  @extend .btn;
  background-color: green;
}
```

I think the extension should happen in the DOM by combining selectors instead of in the Less files.

The other thing is there's nothing "bio" about the `bios-list` right, it's just an `index-list` with a different h3 header style.

It sounds sort of heretical compared to the ways of a few years ago, but I really think "style it in the DOM, extract a component when it reveals itself" is the most maintainable approach nowadays. To me it just comes down to the fact that maintaining markup is usually way easier than maintaining CSS, so the less CSS you write and the less changes you have to make to the CSS, the better, haha...

resources/views/bios/edit.blade.php
besologic: LOL. The commentary in this PR is 🔥

PR #70, opened 2015-05-14
Add conference seeds

adamwathan:
Plus other seed + factory clean up.


database/seeds/ConferencesSeeder.php
```diff
@@ -0,0 +1,42 @@
...
+        $user_ids = collect(User::lists('id'));
+        $conference_names = collect([
+            'MegaAwesomeCon',
+            'SuperPHP',
+            'ActiveRecordCon',
+            'ConCon',
+            'GoodJobFest',
+            'TightenFest',
+            'UltraMegaCon',
+            'ArbysCon',
```
mattstauffer: MMMM
url

database/seeds/ConferencesSeeder.php
```diff
@@ -0,0 +1,42 @@
...
+        ]);
+
+        foreach (range(1, 30) as $i) {
+            $starts_at = $faker->dateTimeBetween('now', '3 years');
```
mattstauffer: This could end up with weird overlaps between CFP and the conference. Do we care?

mattstauffer:
Good to me other than question above

database/seeds/ConferencesSeeder.php
adamwathan: Probably ok for now, I made sure the CFP always starts before the conference, but it could start 3 years before. These could definitely be smarter but it would be more work.

I did all this crap just so I could submit a PR that cleaned up your "+1000 years" sorting code, which I didn't even get around to, haha...

mattstauffer: Bahaha, you didn't like that? :) But no worries, we're gonna delete that code the moment we get to the CFP dashboard anyway.. it's just a quick thing for people I know who will benefit from it right now.

PR #90, opened 2015-06-06
Allow viewing revisions.

### LINKED ISSUE #87: Missing Revisions

mattstauffer:
Fixes #87

Thoughts on the UI? The whole revision UI in general needs clean up, but thught this might be a good start in integrating at least showing revisions

revisions


app/Http/routes.php
```diff
@@ -45,6 +45,8 @@
     Route::get('talks/{id}/delete', 'TalksController@destroy');
     Route::get('conferences/{id}/delete', 'ConferencesController@destroy');
 
+    Route::get('talks/{talkId}/revisions/{revisionId}', 'TalkRevisionsController@show');
```
adamwathan: Should just use `talk-revisions/{id}` or similar, no need to nest, good read: http://weblog.jamisbuck.org/2007/2/5/nesting-resources

_Or_, like you suggested in Slack, `talks/{id}?revision=2r89ce298`, which I think I actually like the looks of better.

mattstauffer: Yah, the more I think about it--especially given how the talk versions controller ended up--i think I prefer query parameyer

Sent from Outlook




On Sat, Jun 6, 2015 at 7:38 AM -0700, "Adam Wathan" <notifications@github.com> wrote:












In app/Http/routes.php:
> @@ -45,6 +45,8 @@
>      Route::get('talks/{id}/delete', 'TalksController@destroy');
>      Route::get('conferences/{id}/delete', 'ConferencesController@destroy');
>  
> +    Route::get('talks/{talkId}/revisions/{revisionId}', 'TalkRevisionsController@show');



Should just use talk-revisions/{id} or similar, no need to nest, good read: http://weblog.jamisbuck.org/2007/2/5/nesting-resources



Or, like you suggested in Slack, talks/{id}?revision=2r89ce298, which I think I actually like the looks of better.

resources/views/talks/show.blade.php
```diff
@@ -9,14 +9,21 @@
             <li><a href="/talks/{{ $talk->id }}">Talk: {{ $current->title }}</a></li>
         </ol>
 
+        @if ($talk->current()->id == $current->id)
```
adamwathan: Is `$current` here the revision currently being displayed, while `$talk->current()` is the most recent revision? I think this could use some naming love.

Could probably also replace these checks with something like `@if ($revision->isCurrent())`

resources/views/talks/show.blade.php
```diff
@@ -9,14 +9,21 @@
...
             <a href="/talks/{{ $talk->id }}/delete" class="btn btn-danger">Delete &nbsp;<span class="glyphicon glyphicon-trash" aria-hidden="true"></span></a>
         </p>
+        @endif
 
         @if ($current === null)
             <p>No revisions yet. @todo Throw exception etc.</p>
         @else
+
+        @if ($talk->current()->id != $current->id)
+            REVISION: {{ $current-> created_at }} | <a href="/talks/{{ $talk->id }}">Return to current revision</a>
```
adamwathan: Can use a named route here?

adamwathan: Also there's a space in the middle of `$current->(HERE)created_at` which is probably gonna make PHP sad.

mattstauffer:
revisions-b

This UI solves all of our current questions, lessens the feature's prominence, solves our current broken UI, and addresses this problem:

--------

I wrote a long, beautiful description for my talk. I submitted it and didn't get in.

I later trimmed it down to be shorter and more technical, out of fear that I had made it too long. 

Even later I realized that I wanted it to be more like my original vision. "How did I word that in my original composition, anyway?", I ask. 

So I go to the talk, and click the revisions button, and go to the revision I want to copy from.

mattstauffer:
I'm gonna merge this version so the live site isn't broken, and then add an issue for us to re-consider the entire UI for revisions.

PR #95, opened 2015-07-20
Add signup and login to home page

mattstauffer:
It's a simple app. It shouldn't take so many clicks to do things.

image


adamwathan:
Looks good! :+1: MURGE!

Have you noticed that DigitalOcean lets you login from the sign up form? Entering your existing credentials here logs you in:

image

I still like having both forms there because very few people would think to try and login using the sign up form, but still kinda neat :+1:

mattstauffer:
@adamwathan Ah, that's clever! I like that a lot.

PR #102, opened 2015-08-24
Fix open cfp filter

### LINKED ISSUE #98: Conference Filters are confusing
So there are two filters that i confuse: "CFP is Open" and "Unclosed CFP" .. they sound like the same thing but "CFP is Open" actually lists CFP's that have closed.

Also `http://symposiumapp.com/conferences?filter=unclosed_cfp&sort=closing_next` does not sort ASC from closes end date, rather DESC it seems.

kayladnls:
Open CFP filter wasn't working correctly, and there was a seemingly duplicate "Unclosed CFP" filter. 

Fixed one, removed the other. 

Fixes #98

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 /vendor
 .env
 .sass-cache/
+/.idea
--- a/app/Http/Controllers/ConferencesController.php
+++ b/app/Http/Controllers/ConferencesController.php
@@ -52,9 +52,6 @@ class ConferencesController extends BaseController
             case 'open_cfp':
                 $conferences = Conference::openCfp()->get();
                 break;
-            case 'unclosed_cfp':
-                $conferences = Conference::unclosedCfp()->get();
-                break;
             case 'all':
                 $conferences = Conference::all();
                 break;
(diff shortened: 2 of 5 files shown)


mattstauffer:
@kayladnls Thanks! The unclosed is actually different than open. It needs renaming, but the difference is:

Open CFP Means you can submit RIGHT NOW.
Unclosed CFP means you can submit either now or in the future; regardless, it hasn't closed yet.

I agree it's super confusing. Merits some thinking. But not sure if I want to drop it yet until we can figure out how to bring it back in.

Thanks for catching the typo here though!

kayladnls:
If you like, I can split the link change into a separate PR?

mattstauffer:
Please--thanks!

On Mon, Aug 24, 2015 at 11:26 AM, Kayla Daniels <notifications@github.com>
wrote:

> If you like, I can split the link change into a separate PR?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/tightenco/symposium/pull/102#issuecomment-134249312>.
>



-- 

*Matt Stauffer*
*Partner + Technical Director*
Tighten Co.
352-682-2765

PR #131, opened 2015-10-25
Add mcrypt extension Composer requirement

elazar:
mcrypt is expected due to [its use](https://github.com/tightenco/symposium/blob/b0392c495faad361f699d0144142dcb4e9cbd030/config/app.php#L83) in the app config file, but this is not indicated when installing Composer dependencies. mcrypt isn't available in all environments (e.g. when installing PHP on OS X via Homebrew, mcrypt must be installed separately from the core PHP distribution). As such, it should be explicitly indicated as a dependency in `composer.json`.


mattstauffer:
:+1:

PR #133, opened 2015-10-25
Add storage/ to .gitignore

elazar:
This change prevents files under the `storage/` directory, such as
log files and SQLite database files, from appearing in git status
output.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 .env
 .sass-cache/
 /.idea
+/storage


mattstauffer:
@elazar If you take a look at the directories under `storage`, they each have manually configured `.gitignore` files. FYI, more common locations for `sqlite` databases are the `database` folder, and then manually adding an explicit `.gitignore` for each.

Thanks!

elazar:
> more common locations for sqlite databases are the `database` folder

This [isn't presently reflected](https://github.com/tightenco/symposium/blob/6b07d5e7979e03a4872051cae420d3682881ca10/config/database.php#L57) in the default configuration.

I'm not sure I'm following the reasoning behind this setup with `.gitignore` files in subdirectories. To my understanding, git doesn't take these into account, only the file in the root directory of the repository. I'm not sure I see the purpose in having separate files if the user will end up having to manually edit them regardless.

mattstauffer:
:shrug: That's fine as a location--again, my teams tend to use `database/`, but I don't really care where we have it for this project. Again, feel free to add it that file explicitly in `storage`'s `.gitignore` if you're using it there; I'm 100% fine with that.

Regarding the `.gitignore` files in subdirectories, they are indeed taken into account. If you don't like them, I'd suggesting taking it up with Laravel, as that's where the current setup comes from. :) However, one context in which it's extremely common is when you want to keep a directory but ignore most or all of its contents; `.gitkeep` is a common convention but actually means nothing to Git; rather, adding a `.gitignore` file in that directory that ignores everything but itself is a very common pattern.

elazar:
Ah, I stand corrected. Fair enough. Was primarily trying to reduce potential friction for new contributors by preventing accidental inclusion of the database file in a commit, assuming the default configuration was used.

> Again, feel free to add it that file explicitly in `storage`'s `.gitignore` if you're using it there; I'm 100% fine with that.

In my local repo, or did you mean for me to submit a PR with that change?

mattstauffer:
PR would be fine! Just add `database.sqlite` to the `storage/.gitignore` file. :) Thanks @elazar!

PR #134, opened 2015-10-25
Fix #122: Add Slides URL field to Talks

### LINKED ISSUE #122: Add slides URL to talk profile
Feature request: OpenCFP (the most common platform for entering data) has a "Slides URL" field, and when used with https://github.com/rdohms/pronto, it would be great if Pronto could grab that field from Symposium and pop it into OpenCFP :)

elazar:


database/migrations/2015_10_25_125123_add_talk_revision_slides.php
```diff
@@ -0,0 +1,31 @@
+<?php
+
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Database\Migrations\Migration;
+
+class AddTalkRevisionSlides extends Migration
```
elazar: This isn't the case for existing database migrations, so I'm assuming it can be ignored.

mattstauffer: Yep :+1: We need to add a `nitpick.json` to ignore the tests dir. I'll do that elsewhere

PR #141, opened 2015-11-12
Redirect failed auth attempt to /log-in.

### LINKED ISSUE #137: Login failure from the front page shows the error in the signin box
image

jwalton512:
Closes #137 

On a failed log in attempt, the user will be routed to /log-in (even from front page) and the error message will display in the appropriate place.


mattstauffer:
Thanks @jwalton512!!  :+1:

PR #145, opened 2015-11-19
Add nitpick configuration

jwalton512:
Set nitpick to ignore some directories that are known to not meet psr-2 (specifically, top level namespace).


mattstauffer:
Thanks @jwalton512!

PR #156, opened 2016-02-12
Readme updates and composer script to run tests.

jonathantorres:


mattstauffer:
Thanks for the contribution! And thanks for cleaning up the indentations.

With regard to the `composer test`, I like that idea, but I think it's probably better to just get folks used to running `./vendor/bin/phpunit`. Maybe I'm just old school?

Either way, though, you're right that that should be in the README.

I'll leave this open for a bit just to see if anyone else gives any feedback on whether I'm just being annoyingly old-fashioend with wanting to do it this way. And either way, thanks Jonathan!

mattstauffer:
What the heck. Let's do it.

mattstauffer:
Thanks @jonathantorres!

Issue #154, opened 2016-01-25
Delete account


nickels:
Delete account from `http://symposiumapp.com/account/delete` results in error page with `Whoops, looks like something went wrong.`

PR #159, opened 2016-02-13
Delete account and relationships. 

jonathantorres:
Solves #154 
- Migration adds cascade on delete constraints on user relationships. 
- Added a test that confirms the fix.

### DIFF
--- a/app/Http/Controllers/AccountController.php
+++ b/app/Http/Controllers/AccountController.php
@@ -94,10 +94,12 @@ class AccountController extends BaseController
 
     public function destroy()
     {
-        $user = Auth::user()->id;
-        $user->delete();
+        $userId = Auth::user()->id;
+        $user = User::find($userId);
 
         Auth::logout();
+
+        $user->delete();
 
         Session::flash('message', 'Successfully deleted account.');
 
--- a/app/Http/Controllers/AccountController.php
+++ b/app/Http/Controllers/AccountController.php
@@ -94,11 +94,9 @@ class AccountController extends BaseController
 
     public function destroy()
     {
-        $userId = Auth::user()->id;
-        $user = User::find($userId);
+        $user = Auth::user();
 
         Auth::logout();
-
         $user->delete();
 
         Session::flash('message', 'Successfully deleted account.');
--- a/tests/AccountTest.php
+++ b/tests/AccountTest.php
@@ -1,0 +1,23 @@
+<?php
+
+use Laracasts\TestDummy\Factory;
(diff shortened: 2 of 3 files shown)


app/Http/Controllers/AccountController.php
```diff
@@ -94,11 +94,13 @@ public function delete()
 
     public function destroy()
     {
-        $user = Auth::user()->id;
-        $user->delete();
+        $userId = Auth::user()->id;
+        $user = User::find($userId);
```
mattstauffer: Wow, that old code looks very broken. But I'm pretty sure we can just have this be `$user = Auth::user()`?

tests/AccountTest.php
```diff
@@ -0,0 +1,23 @@
...
+        $user = Factory::create('user');
+        $this->actingAs($user)
+             ->visit('account/delete')
+             ->press('Yes')
+             ->seePageIs('/')
+             ->see('Successfully deleted account.');
+
+        $this->dontSeeInDatabase('users', [
+            'email' => $user->email,
+        ]);
```
mattstauffer: We'll probably want to test the cascade as well. but this is a great start--thanks!

app/Http/Controllers/AccountController.php
jonathantorres: yeah :+1:

mattstauffer:
Well, I tried adding this to AppServiceProvider@boot:

```php
if (DB::connection() instanceof \Illuminate\Database\SQLiteConnection) {
    DB::statement(DB::raw('PRAGMA foreign_keys=1'));
}
```

... in order to enable foreign keys in SQLite. However, it broken everything; I'm guessing we have tests that insert some data without correct foreign key references. Or something.

Regardless, I wonder whether the better solution isn't just to explicitly delete all of the child/related items by using model events and listening to the deleted event: http://laravel-tricks.com/tricks/cascading-deletes-with-model-events

jonathantorres:
Would you consider using MySQL for the tests? I can help you set that up if you want (in a separate PR of course)

mattstauffer:
@jonathantorres Honestly, I think i'd prefer explicitly deleting in a `deleting` event over switching the stack. I think I'm just ready to commit: That's what I'd like to do with this. :)

jonathantorres:
Sounds good! :+1:

mattstauffer:
@jonathantorres thank you so much for your work on this!

jonathantorres:
Happy to help!

PR #167, opened 2016-05-17
Enhancement: Clarify state when logged in

kayladnls:
The home page in a logged in state never fails to confuse me.  This PR attempts to clarify that a little by: 

- [x] Removing "Sign up" panel when logged in. 
- [x] removing "Log In" panel header when logged in.


mattstauffer:
:+1: Thanks!!

PR #168, opened 2016-05-17
Switch to "Log in" instead of "Sign in" on submit button

kayladnls:
Every where on the site says "Log in" and "Log out" except this one button. 

image


mattstauffer:
:+1: Nice catch--thanks!

PR #171, opened 2016-06-02
Conference end dates must be after start dates

### LINKED ISSUE #170: Able to create a conference with an 'ends_at' before 'starts_at'
I was able to create a conference that had an ending date before the starting date.

As well the `cpf_ends_at` date can be before the `cpf_starts_at` field.

I can put together a PR for this but was curious about the rules in the `CreateConferenceForm`.

It has each validation rule as an array and if I try something like this:

```
        'ends_at' => ['date|after:starts_at'],
```
I get a:
```
BadMethodCallException" matches expected exception "Symposium\Exceptions\ValidationException". Message was: "Method [validateDate|after] does not exist.
```

However this works just fine:
```
        'ends_at' => 'date|after:starts_at',
```

Was there a reason for each rule value being an array like that? Maybe I am missing something obvious? :confused:

bmitch:
Fixes #170

### DIFF
--- a/app/Services/CreateConferenceForm.php
+++ b/app/Services/CreateConferenceForm.php
@@ -12,9 +12,9 @@ class CreateConferenceForm
         'description' => ['required'],
         'url' => ['required'],
         'starts_at' => ['date'],
-        'ends_at' => ['date'],
+        'ends_at' => ['date', 'after:starts_at'],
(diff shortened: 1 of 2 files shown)


mattstauffer:
Thanks @bmitch!

PR #173, opened 2016-06-04
CFP dates must be before conference start date

bmitch:
Figured this validation would make sense. CFP would be before the start of a conference.


mattstauffer:
Thanks! :)

bmitch:
No problem. I also noticed you can make a conference in the past. Is this intentional or should conferences only be in the future? I can add another validation on that if you like.

mattstauffer:
@bmitch hm, I haven't totally thought that through, to be honest. Let's keep it as a possibility for now, but thanks!

PR #179, opened 2016-07-27
Allow Talks to be archived

myleshyson:
Started working on this as a response to https://github.com/tightenco/symposium/issues/126

I’m new to php and Laravel and it’s my first time contributing to an open source project. Would love feedback on how functional the code is.


.env.example
```diff
@@ -1,27 +0,0 @@
-APP_ENV=local
```
mattstauffer: Please un-delete `.env.example`; it's a useful file. :)

.env.example
```diff
@@ -1,27 +0,0 @@
...
-
-TWITTER_CONSUMER_KEY=
-TWITTER_CONSUMER_SECRET=
-TWITTER_ACCESS_TOKEN=
-TWITTER_ACCESS_TOKEN_SECRET=
-
-SLACK_ENDPOINT=
-
-CAPTCHA_PUBLIC=
-CAPTCHA_PRIVATE=
```
mattstauffer: Please` .gitignore` your `.tags` file

app/Http/Controllers/TalksController.php
```diff
@@ -102,7 +85,7 @@ public function store()
     public function edit($talkId)
     {
         try {
-            $talk = Auth::user()->talks()->findOrFail($talkId);
+            $talk = Auth::user()->talks()->withTrashed()->findOrFail($talkId);
```
mattstauffer: Hm, soft deletes and archive aren't usually the same thing. I get the idea of using soft deleted entries as archived talks, but that's not really how they're intended to be used, and I think this could lead to a lot of confusion. I'd prefer a "is_deleted" flag or status of some sort that we can then do a global scope on, or something similar.

app/Http/Controllers/TalksController.php
```diff
@@ -157,10 +141,72 @@ public function show($id)
...
+    public function archiveIndex()
+    {
+        list($talks, $sorting_talk) = $this->sortTalks('archive');
+
+        return View::make('talks.archive')
+          ->with('talks', $talks)
+          ->with('sorting_talk', $sorting_talk);
+    }
+
+  /**
```
mattstauffer: Looks like you're having indentation issues here

app/Http/Controllers/TalksController.php
```diff
@@ -157,10 +141,72 @@ public function show($id)
...
+        $sorting_talk = $this->sorting_talks;
+
+        if ($type == 'archive') {
+            $talks = Auth::user()->talks()->onlyTrashed()->get();
+        } else {
+            $talks = Auth::user()->talks;
+        }
+
+        switch (Input::get('sort')) {
+        case 'date':
```
mattstauffer: Please fix all these Nitpick errors :)

app/Http/Controllers/TalksController.php
```diff
@@ -157,10 +141,72 @@ public function show($id)
...
+   * @param  string $type Whether the index is for archived talks or active talks
+   * @return array       Returns array of variables.
+   */
+    private function sortTalks($type = 'active')
+    {
+        $bold_style = 'style="font-weight: bold;"';
+
+        $sorting_talk = $this->sorting_talks;
+
+        if ($type == 'archive') {
```
mattstauffer: I think I'd just pass the `$talks` variable into `sortTalks` instead of passing the type.

app/Http/Controllers/TalksController.php
```diff
@@ -157,10 +141,72 @@ public function show($id)
...
+        case 'alpha':
+            // Pass through
+        default:
+            $sorting_talk['alpha'] = $bold_style;
+            $talks = $talks->sortBy(function ($talk) {
+              return strtolower($talk->current()->title);
+            });
+            break;
+    }
+        return array($talks, $sorting_talk);
```
mattstauffer: You can use short array syntax: `return [$talks, $sorting_talk]`

app/models/Talk.php
```diff
@@ -1,9 +1,17 @@
...
+use Illuminate\Database\Eloquent\SoftDeletes;
+
 class Talk extends UuidBase
 {
+    use SoftDeletes;
+
     protected $table = 'talks';
 
+    protected $dates = [
+      'delted_at'
```
mattstauffer: Typo here (but that won't matter when we switch to archive instead of soft delete)

resources/assets/sass/app.scss
```diff
@@ -243,7 +243,7 @@ legend {
 
 .page-meta-buttons {
     .btn {
-        float: right;
+        display: block;
```
mattstauffer: hm, what's this for?

resources/views/partials/talk-in-list.blade.php
```diff
@@ -1,12 +1,25 @@
...
+  </a>
     <a href="{{ route('talks.delete', ['id' => $talk->id]) }}" class="btn btn-xs btn-danger">
         <span class="glyphicon glyphicon-trash" aria-hidden="true"></span>
         Delete
     </a>
-    <a href="{{ route('talks.edit', ['id' => $talk->id]) }}" class="btn btn-xs btn-primary">
-        <span class="glyphicon glyphicon-edit" aria-hidden="true"></span>
-        Edit
-    </a>
+    @if (Route::is('talks.archive'))
```
mattstauffer: I think you'd do better to switch this based on a property of the talk itself; e.g. `$talks->isArchived()`

resources/views/talks/archive.blade.php
```diff
@@ -0,0 +1,28 @@
...
+@section('content')
+
+    <div class="container body">
+        <div class="row">
+            <div class="col-md-8 col-md-push-2">
+
+                <h2 class="page-title">Archive</h2>
+                <a href="/talks" class="btn btn-default btn-xs">Back</a>
+
+                <p class="list-sort">Sort: <a href="{{ route('talks.archive') }}?sort=alpha"{{ $sorting_talk['alpha'] }}>Title</a> | <a
```
mattstauffer: These lines are a bit long. I'd probably indent a bit differently here

resources/views/talks/index.blade.php
```diff
@@ -8,6 +8,8 @@
                 <a href="{{ route('talks.create') }}" class="create-button">Create Talk &nbsp;<span class="glyphicon glyphicon-plus" aria-hidden="true"></span></a>
 
                 <h2 class="page-title">My Talks</h2>
+                <a href="{{ route('talks.archive') }}" class="btn btn-default btn-xs">Archive</a>
+
```
mattstauffer: No need for this extra line break

app/Http/Controllers/TalksController.php
spuds10: Hey Matt, sounds great. Just to clarify, do you mean for example adding a boolean column called archived, then just making a scope to reference all the articles where that is true?

app/Http/Controllers/TalksController.php
spuds10: Sure thing! Im unfamiliar with nitpick. So does it just show the portions that are indented incorrectly. For example, for the code above, is it saying that this whole block is indented incorrectly?

app/Http/Controllers/TalksController.php
mattstauffer: Yep, something like that!

resources/assets/sass/app.scss
spuds10: So originally the show talk page had the buttons stacked. When I added an archive button, it broke how the buttons stacked. It ended up looking like this
symposium for speakers conferences

To fix this I made the buttons block and took off the float right. So it looks like this now
symposium for speakers conferences

myleshyson:
@mattstauffer Alright I worked on those changes and fixed the nitpick errors. Let me know what you think!

.gitignore
```diff
@@ -1,5 +1,6 @@
 /node_modules
 /vendor
 .env
+.tags
```
mattstauffer: drop `.tags` and `.idea`; these are good for your own local `.gitignore`

app/Http/Controllers/TalksController.php
```diff
@@ -32,7 +32,8 @@ class TalksController extends BaseController
 
     public function index()
     {
-        list($talks, $sorting_talk) = $this->sortTalks();
+        $talks = Auth::user()->talks()->active()->get();
```
mattstauffer: Maybe consider:

```php
list($talks, $sorting_talk) = $this->sortTalks(
    Auth::user()-talks()->active()->get()
);
```

app/Http/Controllers/TalksController.php
```diff
@@ -141,72 +142,70 @@ public function show($id)
...
         Session::flash('message', 'Successfully deleted talk.');
 
         return Redirect::to('talks');
     }
 
-    public function softDelete($id)
+    public function archiveIndex()
     {
-        Auth::user()->talks()->findOrFail($id)->delete();
+        $talks =  Auth::user()->talks()->archived()->get();
```
mattstauffer: Same refactor as above

app/Http/Controllers/TalksController.php
```diff
@@ -141,72 +142,70 @@ public function show($id)
...
+
+        return View::make('talks.archive')
+          ->with('talks', $talks)
+          ->with('sorting_talk', $sorting_talk);
+    }
+
+    public function archive($id)
+    {
+        $talk = Auth::user()->talks()->findOrFail($id);
+        $talk->is_archived = true;
```
mattstauffer: Imagine if the syntax is:

```php
Auth::user()->talks()->findOrFail($id)->archive();
```

this would require adding an `archive()` and probably an `unarchive()` method to the model

app/Http/Controllers/TalksController.php
```diff
@@ -141,72 +142,70 @@ public function show($id)
...
-        list($talks, $sorting_talk) = $this->sortTalks('archive');
-
-        return View::make('talks.archive')
-          ->with('talks', $talks)
-          ->with('sorting_talk', $sorting_talk);
-    }
-
   /**
    * Sorts through talks and returns an array of variables. Access the variables using the list() construct.
    * @param  string $type Whether the index is for archived talks or active talks
```
mattstauffer: param is wrong; docblocks suck, BUT this has a weird return so maybe doc block could help me (?) if it were more clear

app/Http/Controllers/TalksController.php
```diff
@@ -141,72 +142,70 @@ public function show($id)
...
    * @return array       Returns array of variables.
    */
-    private function sortTalks($type = 'active')
+    private function sortTalks($talks)
     {
         $bold_style = 'style="font-weight: bold;"';
 
         $sorting_talk = $this->sorting_talks;
```
mattstauffer: Why don't you just make these changes to `$this->sorting_talks`, and then just return the `$talks` variable? No need for list,

app/Http/Controllers/TalksController.php
```diff
@@ -141,72 +142,70 @@ public function show($id)
...
-
   /**
    * Sorts through talks and returns an array of variables. Access the variables using the list() construct.
    * @param  string $type Whether the index is for archived talks or active talks
    * @return array       Returns array of variables.
    */
-    private function sortTalks($type = 'active')
+    private function sortTalks($talks)
     {
         $bold_style = 'style="font-weight: bold;"';
```
mattstauffer: Please get rid of `$Bold_style`. Instead, if a sort is chosen, set it to be `true` instead of `false`; then in the template, echo this same style *if* the given sort is `true`

app/models/Talk.php
```diff
@@ -47,8 +43,23 @@ public function getRevisionsAttribute()
         return $this->revisions()->getQuery()->orderBy('created_at', 'desc')->get();
     }
 
+    public function isArchived()
+    {
+        return $this->attributes['is_archived'];
```
mattstauffer: double check that this is always going to return a boolean; i don't know whether it will get cast on its way into `attributes`. Could you just do `$this->is_archived`

resources/views/talks/archive.blade.php
```diff
@@ -5,12 +5,12 @@
...
         <div class="row">
             <div class="col-md-8 col-md-push-2">
-
                 <h2 class="page-title">Archive</h2>
                 <a href="/talks" class="btn btn-default btn-xs">Back</a>
-
-                <p class="list-sort">Sort: <a href="{{ route('talks.archive') }}?sort=alpha"{{ $sorting_talk['alpha'] }}>Title</a> | <a
-                            href="{{ route('talks.archive') }}?sort=date"{{ $sorting_talk['date'] }}>Date</a></p>
+                <p class="list-sort">Sort:
+                    <a href="{{ route('talks.archived.index') }}?sort=alpha"{{ $sorting_talk['alpha'] }}>Title</a> |
```
mattstauffer: Not sure about best solution, but here's one:

```html
<a href="{{ route('talks.archived.index') }}?sort=alpha" @if ($sorting_talk['alpha'])style="yada"@endif>Title</a> 
```

resources/views/talks/index.blade.php
```diff
@@ -5,14 +5,15 @@
...
+                </a>
                 <h2 class="page-title">My Talks</h2>
-                <a href="{{ route('talks.archive') }}" class="btn btn-default btn-xs">Archive</a>
-
-
-                <p class="list-sort">Sort: <a href="{{ route('talks.index') }}?sort=alpha"{{ $sorting_talk['alpha'] }}>Title</a> | <a
-                            href="{{ route('talks.index') }}?sort=date"{{ $sorting_talk['date'] }}>Date</a></p>
+                <a href="{{ route('talks.archived.index') }}" class="btn btn-default btn-xs">Archive</a>
+                <p class="list-sort">Sort:
+                    <a href="{{ route('talks.index') }}?sort=alpha"{{ $sorting_talk['alpha'] }}>Title</a> |
```
mattstauffer: same thing here

app/Http/Controllers/TalksController.php
```diff
@@ -25,19 +25,21 @@ class TalksController extends BaseController
         'slides.url' => 'Slides URL must contain a valid URL',
     ];
 
-    protected $sorting_talks = [
+    protected $sorting_style = [
         'date' => '',
         'alpha' => '',
     ];
```
myleshyson: I left it as strings  so that in the sortTalks method, it sets it to a classname. I also renamed it to represent that a little better. That way in the blade view, it looks like this.
```
<a href="{{ route('talks.index') }}?sort=alpha" class="{{ $sorting_style['alpha'] }}">Title</a>
```

myleshyson:
@mattstauffer Alright made those changes, let me know what you think!

resources/assets/js/app.js
```diff
@@ -31,7 +31,7 @@ $(function() {
     });
 
     $('[data-dismiss=timeout]').each(function() {
-        var timeout_len = 2000,
+        var timeout_len = 1000,
```
mattstauffer: :) Glad you remembered we talked about this! however, a shorter timeout means some folks won't be able to read it. Rather, we need to make it dismissable. It feels like that wasn't an option when I was running into it; could you re-up the timeout, and check whether the dismiss is working? Thanks!

app/Http/Controllers/TalksController.php
mattstauffer: I like that you cleaned it up--it's definitely better than having style in a PHP var--but we still have the same problem of having style information in our PHP, which is not preferable. If we use a boolean, yes, it clutters up our views, but it makes this array a representation of "which is being sorted by" instead of "classes to apply", which is a much better separation.

To be honest.. one of the interesting questions is, why is this even an array? It feels like there could just be a "sortingBy" property which is set to the name of the field we're sorting by, and then just call it a day.

app/Http/Controllers/TalksController.php
```diff
@@ -25,19 +25,21 @@ class TalksController extends BaseController
...
 
     public function index()
     {
-        $talks = Auth::user()->talks()->active()->get();
-        list($talks, $sorting_talk) = $this->sortTalks($talks);
+        $talks = $this->sortTalks(
+            Auth::user()->talks()->active()->get()
+        );
+        $sorting_style = $this->sorting_style;
```
mattstauffer: You don't need a temp variable. Just pass `$this->sorting_style` directly to the `with` method.

app/Http/Controllers/TalksController.php
```diff
@@ -151,61 +153,51 @@ public function destroy($id)
 
     public function archiveIndex()
     {
-        $talks =  Auth::user()->talks()->archived()->get();
-        list($talks, $sorting_talk) = $this->sortTalks($talks);
+        $talks = $this->sortTalks(
+            Auth::user()->talks()->archived()->get()
+        );
+
+        $sorting_style = $this->sorting_style;
```
mattstauffer: Same--no need for temp var here

app/Http/Controllers/TalksController.php
```diff
@@ -151,61 +153,51 @@ public function destroy($id)
...
-          ->with('sorting_talk', $sorting_talk);
+          ->with('sorting_style', $sorting_style);
     }
 
     public function archive($id)
     {
-        $talk = Auth::user()->talks()->findOrFail($id);
-        $talk->is_archived = true;
-        $talk->save();
+        $talk = Auth::user()->talks()->findOrFail($id)->archive();
```
mattstauffer: Don't even need to save it to `$talk` variable, because you're not going to use it anywhere. :) So fresh. So clean.

app/Http/Controllers/TalksController.php
```diff
@@ -151,61 +153,51 @@ public function destroy($id)
...
         return Redirect::to('talks');
     }
 
-    public function restore($talkId)
+    public function restore($id)
     {
-        $talk = Auth::user()->talks()->findOrFail($talkId);
-        $talk->is_archived = false;
-        $talk->save();
+        $talk = Auth::user()->talks()->findOrFail($id)->restore();
```
mattstauffer: No need for temp var.

resources/assets/sass/app.scss
```diff
@@ -69,6 +69,10 @@ legend {
     width: inherit;
 }
 
+.bold {
```
mattstauffer: If it's a utility class, we should probably prefix it with something like `u-`. I don't know if we have any other utility classes in Symposium we can use as reference points, though, but that's my general way of thinking. `.u-bold {}`

mattstauffer:
@myleshyson Changes reviewed :+1:

myleshyson:
@mattstauffer Alright! made some new changes :) So for the blade view, I made a custom blade directive that takes in the sorted talks and what your passing it against. Let me know what you think!

app/Http/Controllers/TalksController.php
```diff
@@ -163,4 +146,52 @@ public function destroy($id)
...
+
+        return Redirect::to('archive');
+    }
+
+    private function sortTalks($talks)
+    {
+        switch (Input::get('sort')) {
+            case 'date':
+                $this->sorted_by = 'date';
+                $talks = $talks->sortByDesc('created_at');
```
mattstauffer: Can we just `return $talks->sortByDesc` and same below, instead of setting a temp var and then returning?

app/Http/Controllers/TalksController.php
```diff
@@ -25,35 +25,17 @@ class TalksController extends BaseController
         'slides.url' => 'Slides URL must contain a valid URL',
     ];
 
-    protected $sorting_talks = [
-        'date' => '',
-        'alpha' => '',
-    ];
+    protected $sorted_by;
```
mattstauffer: Even though it's not necessary, let's set this equal to `alpha` here

app/models/Talk.php
```diff
@@ -1,15 +1,19 @@
 <?php
 
+
```
mattstauffer: please drop all these extra lines

app/models/Talk.php
```diff
@@ -1,15 +1,19 @@
...
     protected $guarded = [
         'id'
     ];
 
     protected $casts = [
-        'public' => 'boolean'
+        'public' => 'boolean',
+        'is_archived' => 'boolean'
```
mattstauffer: comma after `'boolean'`

resources/views/partials/header.blade.php
```diff
@@ -1,3 +1,4 @@
+
```
mattstauffer: drop this line plz

resources/views/talks/archive.blade.php
```diff
@@ -0,0 +1,28 @@
+@extends('layout')
+
+@section('content')
+
```
mattstauffer: drop this line plz

resources/views/talks/archive.blade.php
```diff
@@ -0,0 +1,28 @@
...
+    <div class="container body">
+        <div class="row">
+            <div class="col-md-8 col-md-push-2">
+                <h2 class="page-title">Archive</h2>
+                <a href="/talks" class="btn btn-default btn-xs">Back</a>
+                <p class="list-sort">Sort:
+                    <a href="{{ route('talks.archived.index') }}?sort=alpha" class="@sorted($sorted_by, 'alpha')">Title</a> |
```
mattstauffer: you can pass the sort = alpha into route like this:

```php
{{ route('talks.archived.index', ['sort' => 'alpha']) }}
```

mattstauffer:
@myleshyson OK, reviewed! Once you make these changes ping me and I'll test it out locally and then we can get this pushed!

myleshyson:
@mattstauffer Hey Matt! Cleaned up those lines.

mattstauffer:
OK, let's do this. :) Gonna merge it in a few.

myleshyson:
Woo! My bad about the use statements. I'll know that for next time :)

rdohms:
@mattstauffer @myleshyson is this also implemented in API? cause by default it still shows all talks archived and unarchived alike.

Is there a flag i can pass to get one, other or both?

mattstauffer:
@rdohms Just pushed an update where archived are excluded from API; future work will allow you to bring them back or request only them. #202

Issue #151, opened 2015-12-18
HiRes Profile picture


rdohms:
Allow the upload of higher res profile pictures and add a link to "full size" download on the profile page.

This allows us to send the profile page to conferences and they can grab pictures and bio's from there.

mattstauffer:
:+1:

mattstauffer:
Completed with #182

PR #181, opened 2016-08-11
Created upload of profile picture.

michaelrtm:
This PR is for an upload Profile Picture as per #116 
The image manipulation is through a new composer package "intervention/image" and is handled in an event.
As per #151 there is also a hires version of the image that is accessible by clicking the profile picture on the public profile view.


michaelrtm:
Before you potentially merge this I'd like to add perhaps a file size restriction (5mb?), and then delete the original uploaded image as well.

michaelrtm:
I'm rethinking the event firing - I don't like that the file name is being updated in the model potentially before the listener has executed

michaelrtm:
I think I'm there

PR #182, opened 2016-08-13
Add Profile Picture

michaelrtm:
This PR is for an upload Profile Picture as per #116 
Image manipulation is through a new composer package "intervention/image" and is handled in an event.
As per #151 there is also a hires version of the image that is accessible by clicking the profile picture on the public profile view.

Sorry for stuffing around on the previous pull request. (I've never really done this before :) )


app/Http/Controllers/AccountController.php
```diff
@@ -74,6 +79,12 @@ public function update(Request $request)
         $user->allow_profile_contact = $request->get('allow_profile_contact');
         $user->profile_intro = $request->get('profile_intro');
         $user->profile_slug = $request->get('profile_slug');
+
+        if ($request->file('profile_picture')) {
+            $image = $request->file('profile_picture');
+            Event::fire(new ProfilePictureUpdated($user, $image->getRealPath(), $image->getClientOriginalExtension()));
```
mattstauffer: Are we sure that handling file uploads in a queued event will work? I don't know how long uploaded files stick around...

app/Listeners/UpdateProfilePicture.php
```diff
@@ -0,0 +1,55 @@
...
+
+    /**
+     * Handle the event.
+     *
+     * @param  ProfilePictureUpdated  $event
+     * @return void
+     */
+    public function handle(ProfilePictureUpdated $event)
+    {
+        $filename = time() . '.' . $event->image_ext;
```
mattstauffer: Could we instead use `$image->hashName()` so we don't have any possible collisions?

app/Listeners/UpdateProfilePicture.php
```diff
@@ -0,0 +1,55 @@
...
+
+        //Make hires image
+        Image::make($event->image)
+            ->fit(1250, 1250, function ($constraint) {
+                $constraint->upsize();
+            })
+            ->save(public_path('img/profile_pictures/hires/' . $filename));
+
+        if ($event->user->profile_picture != "missing") {
+            File::delete(public_path('img/profile_pictures/'.$event->user->profile_picture));
```
mattstauffer: Can we update this to use `Storage` instead of `File` so we can move this storage to cloud if we'd like?

app/Listeners/UpdateProfilePicture.php
```diff
@@ -0,0 +1,55 @@
...
+            ->save(public_path('img/profile_pictures/' . $filename));
+
+        //Make hires image
+        Image::make($event->image)
+            ->fit(1250, 1250, function ($constraint) {
+                $constraint->upsize();
+            })
+            ->save(public_path('img/profile_pictures/hires/' . $filename));
+
+        if ($event->user->profile_picture != "missing") {
```
mattstauffer: Using `'missing'` as the default feels off. Can we use `null` instead?

database/migrations/2016_08_13_102611_add_profile_picture_to_users.php
```diff
@@ -0,0 +1,31 @@
...
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('users', function (Blueprint $table) {
+            $table->string('profile_picture')->default('missing.png');;
```
mattstauffer: I don't like the string name here, and you also have two `;` at the end

mattstauffer:
Thanks for this! I've added some notes inline. Two big issues that didn't come across in the notes:

1. This breaks people who currently are relying on Gravatar. A good solution would provide Gravatar as a fallback.
2. The original plan was to have multiple profile pictures (https://github.com/tightenco/symposium/issues/2) but unfortunately that info wasn't moved to the later (https://github.com/tightenco/symposium/issues/116). I'm thinking we can just allow for a single picture for now and address adding more profile pictures later.

Thanks!

app/Http/Controllers/AccountController.php
michaelrtm: I know that methods are supposed to have a single responsibility, that's why I've put it in an event. 

Should I keep the manipulation in the event but move the file to storage in the method, remove the event and keep it all in the method, or remove the ShouldQueue?

mattstauffer: Hm, good question. Since what we're really doing is saying "upload this picture", then it's more like a Job (and one which we should disable queuing for).

database/migrations/2016_08_13_102611_add_profile_picture_to_users.php
michaelrtm: string name, as in _profile_pictue_ or that it defaults to missing?
In the version I'm working on, it defaults to null but I've left it as _profile_picture_

app/Http/Controllers/AccountController.php
michaelrtm: Kept as event/listener, queue removed

app/Listeners/UpdateProfilePicture.php
michaelrtm: done

app/Listeners/UpdateProfilePicture.php
michaelrtm: Done. Will need symlinks on server to use local - set default to 'public'

michaelrtm:
Gravatar is now a fallback for the nav icon and the public profile page. If a user doesn't have a profile picture uploaded, an alert will show explaining that Gravatar is being used and they should uploaded their own image.

Now that `Storage` is being used instead of `File`, symlinks will need to be setup from `/storage/app/public` to `/public/storage` if you are going to keep local storage

Sorry if this conflicts with my last PR. I only discovered branching after starting this feature. :)

mattstauffer:
@michaelrtm Thank you!! :)

PR #186, opened 2016-08-29
Added Vagrant-ansible support for dev environment

jeroenvdgulik:
Signed-off-by: Jeroen v.d. Gulik <jeroen@isset.nl>


mattstauffer:
Thanks for the PR, @jeroenvdgulik! :)

Is there a reason for setting up a unique Ansible setup here rather than just saying "use Homestead"? That's what we tend to do on our projects and it helps keep our projects free from more complicated tooling like this, but I want to be open to other options.

Thanks!

jeroenvdgulik:
I am unfamiliar with Homestead nor the Laravel Eco System. So I set this up locally, and thought I'd share. I find it more intuitive to just `vagrant up` a project, but that's probably because I'm unfamiliar with Homestead.

mattstauffer:
@jeroenvdgulik Ah, I assumed since you're at Laracons and have been an IRC OG since before me that you were familiar with all this stuff. Thank you so much for the work you put into this contribution! However, it's easier for me to not have to take on the burden of maintaining this setup, and instead just to recommend folks use Homestead. I *really* appreciate your work.

rdohms:
May I recommend a middle ground?

Put this setup in a new repo called symposium-ansible the has this repo as a submodule and add a link in the readme.

This way we do not lose the work done, plus people like me who do not have, nor wish to install homestead can use the other repo to setup.

mattstauffer:
@rdohms That's a fine idea--I'm not interested in hosting that repo, but if someone else wants to I'll be happy to point to it as an alternative to Homestead if anyone doesn't like Homestead. :)

PR #187, opened 2016-08-29
Added Markdown support for Abstract/Description of Talk

jeroenvdgulik:
MARKDOWN FOR WHAT

Markdown for What

Signed-off-by: Jeroen v.d. Gulik jeroen@isset.nl

### DIFF
--- a/app/TalkRevision.php
+++ b/app/TalkRevision.php
@@ -30,6 +30,11 @@ class TalkRevision extends UuidBase
         return '/talks/' . $this->talk->id . '/?revision=' . $this->id;
     }
 
+    public function getDescription()
+    {
+        return $this->htmlize($this->getAttribute('description'), false);
+    }
+
     public function getHtmledDescription()
     {
         return $this->htmlize($this->getAttribute('description'));
@@ -40,12 +45,12 @@ class TalkRevision extends UuidBase
         return $this->htmlize($this->getAttribute('organizer_notes'));
     }
 
-    private function htmlize($string)
+    private function htmlize($string, $changeNewLineToBR = true)
     {
         if ($string == '') {
             return '<i>(empty)</i>';
         }
 
-        return str_replace("\n", "<br>", $string);
+        return $changeNewLineToBR ? str_replace("\n", "<br>", $string) : $string;
     }
 }
--- a/composer.json
+++ b/composer.json
@@ -18,6 +18,7 @@
     ],
     "type": "project",
     "require": {
+        "php": ">=5.6.4",
         "laravel/framework": "5.3.*",
         "rexxars/joindin-client": "^0.0.1",
         "guzzlehttp/guzzle": "~5.3|~6.0",
@@ -30,10 +31,10 @@
         "recaptcha/php5": "^2.0",
         "bugsnag/bugsnag-laravel": "^2.0",
         "intervention/image": "^2.3",
-        "php": ">=5.6.4",
         "pragmarx/firewall": "^1.0",
         "laravel/scout": "^2.0",
-        "algolia/algoliasearch-client-php": "^1.14"
+        "algolia/algoliasearch-client-php": "^1.14",
+        "andreasindal/laravel-markdown": "^1.0"
     },
     "require-dev": {
         "phpunit/phpunit": "~5.0",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
         "This file is @generated automatically"
     ],
-    "content-hash": "13cded881836c244c21806dfaf14e1ad",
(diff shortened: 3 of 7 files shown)


mattstauffer:
@jeroenvdgulik Awesome, thanks! I'll take a look. The first few things that come to my mind here:

- That GIF is terrifying
- oh my, that GIF is terrifying
- let's make sure that we're getting the correct output in the API and describing what's going on; i would guess we may end up having to output both the markdown'ed and the non-markdown'ed versions, but maybe we just output the text as-is and that's it?
- I think we will want to clean up the methods around `getHTMLDescription` or whatever a little bit
- We'll need to make sure that this doesn't changes folks' expectations of how line breaks work, etc.. my guess is it won't but it'll be worth taking a look at.

Thanks!

jeroenvdgulik:
> let's make sure that we're getting the correct output in the API and describing what's going on; i would guess we may end up having to output both the markdown'ed and the non-markdown'ed versions, but maybe we just output the text as-is and that's it?

This is why I added my own method instead of re-using the available methods. Could not tell how that would influence API.

> We'll need to make sure that this doesn't changes folks' expectations of how line breaks work, etc.. my guess is it won't but it'll be worth taking a look at.

Markdown does support this natively I think, but you need to replace the `\n` with 2 spaces and then `\n`

jeroenvdgulik:
@mattstauffer is there anything I can do to currently to make this PR more acceptable?

mattstauffer:
@jeroenvdgulik hey Jeroen, thanks for followiing up. Honestly, I just need a half hour some time to dig through what the UI feels like, how it works with pre-existing data, what decisions I should make with regard to the API, etc, and I have been really busy this week with the new baby, etc.

I'll get to this as soon as I can; it's open i a browser tab which means I'm reminded of it every time I think to myself, "Matt, you have too many browser tabs open." :)

jeroenvdgulik:
No worries man! Just wanted to help out if you were swamped :)

mattstauffer:
@rdohms What do you think about how you as a consumer would expect Markdown-formatted descriptions to be output? My guess is that you'd say "I don't care; in the end it's up to them how it's going to end up in the final form", but I just want to double check.

mattstauffer:
Reviewed the UI. Feels good and clean.

Concerns:
- [ ] API output; waiting on @rdohms' thoughts
- [ ] Linebreaks do not output the same; this is the same talk before and after markdown support:
    image
- [ ] I'll still want to refactor the methods a bit

Once we can solve these three this is good to go!

jeroenvdgulik:
Re: LineBreaks

We could just add the <br> and run it through markdown as it wil ignore html tags. Output should then be the same.

mattstauffer:
@rdohms hey could you take a look at this for API concerns?

@jeroenvdgulik sorry for the long, long, long delay. I'm taking some time on Symposium today and hopefuly will be able to make some progress here. :)

jeroenvdgulik:
No worries man, if there's anything I can do to help, just let me know

rdohms:
sorry folks, missed this. 
@mattstauffer i think you have it right, as an API consumer i expect whatever comes out to come my way. As long as its raw markdown and not rendered html, we good.

imjohnbon:
@jeroenvdgulik @mattstauffer 

Couple of notes here as I've been reviewing this:

1. I was getting a breaking error from the markdown package on pages where it tried to render markdown. I realized that the composer package currently contains `"andreasindal/laravel-markdown": "1.0"`, so it's not pulling in the latest version (`1.1.6` I believe) with some bug fixes. We should probably update that to be `"andreasindal/laravel-markdown": "^1.0"` so we the latest fixes are included.

2. I know we were hesitant to change/remove the `getHtmledDescription()` method because of the potential effects it could have on the API, hence the new `getDescription()` method. However, on my global search of the codebase I realized `getHtmledDescription()` is used only once, and it's on the `talks/show.blade.php` page. So it's not even used in the API. Given that, we can probably just delete it entirely in favor of the new `getDescription()` method which is better named now that we're not returning just HTML.

3. The spacing is definitely different, as Matt noted, but I think we have a solid option here. The markdown package comes with a `markdown.php` config file. The last option in that config file is called `breaks` and it allows you to override how line breaks are handled. If you set it to true (it defaults to false), it uses "github flavored" line breaks. This essentially means it wraps everything in a `<p>` tag if there is a single line break. (Note: naturally we can't edit the `markdown.php` file from the package directly, we'd need to create our own `config/markdown.php` override). With that setting updated, the line breaks are closer to how they look now:

screen shot 2017-06-20 at 12 52 16 pm

Yes, there is a tiny bit of difference when there is a double line break (because the HTML version adds 2 `<br>` tags and the markdown version wraps it in a `<p>` tag, but I personally think it's pretty negligible.

jeroenvdgulik:
I rebased from Master

josecanhelp:
Closing since Markdown support was added in #250.

PR #188, opened 2016-10-21
Bugfix: Bug introduced in aa59503a returning associative array

Xethron:
Laravel's Illuminate\Support\Collection::sortBy() preserves the keys,
and therefore will always return an associative array, even when it
used to be sequential. If you json_encode the result, you might
expect an array, but get an object instead. This commit fixes that
issue, by discarding the array keys, and fetching only its values.
This fixes an issue with vue.js not being able to .forEach over an
object.

### DIFF
--- a/app/User.php
+++ b/app/User.php
@@ -30,7 +30,7 @@ class User extends Authenticatable
     {
         return $this->talks()->get()->sortBy(function ($talk) {
             return strtolower($talk->current()->title);
-        });
+        })->values();
     }
 
     public function bios()


mattstauffer:
Oh snap! Great catch on that! Thank you!!!!!

mattstauffer:
@Xethron Wrote a test to cover it: https://github.com/tightenco/symposium/commit/f75aa23a9d2a1fb77f9d991d1dd622325dff9223

Xethron:
@mattstauffer Awww, was going to do that tonight :P But that looks good!
So, what else needs done? I have 3 more hacktoberfest PR's to do ;) haha

Honestly, this was super easy to setup and fix! Like, wow! Great work with the migrations/seeds/everything. Took like 5 minutes and I had everything ready for testing.

Xethron:
I did however login as Matt, I hope that's allowed ;)

mattstauffer:
@Xethron haha, whoops!

Glad to hear it--that's super encouraging! 

This is a bit more complex, but if you're looking for a task to take, this is one you might like:

https://github.com/tightenco/symposium/issues/169

or this:

https://github.com/tightenco/symposium/issues/166

Or really anything here with the "help wanted" label:

https://github.com/tightenco/symposium/issues

PR #189, opened 2016-10-23
Allow conferences to be one day conferences

Xethron:
The conference ends_at and starts_at should be allowed to be the same day. Laravel's Validator class currently doesn't support this, so the validator was extended to add a 'onOrAfter' validation rule.


mattstauffer:
:+1: Thanks!

PR #190, opened 2016-10-25
Move all models into same folder/namespace

Xethron:
The User model was located under app/ with a namespace of App, while the rest of the models where under app/models/ and had no namespace.

This patch moves all the models into the App\Models namespace and app/Models directory to follow PSR4 Autoloading.

PhpStorm also had a go at removing unused use statements.

### DIFF
--- a/app/ApiResources/Bio.php
+++ b/app/ApiResources/Bio.php
@@ -1,6 +1,6 @@
 <?php namespace App\ApiResources;
 
-use Bio as EloquentBio;
+use App\Models\Bio as EloquentBio;
 use Illuminate\Contracts\Support\Arrayable;
 
 class Bio implements Arrayable
--- a/app/ApiResources/Bio.php
+++ b/app/ApiResources/Bio.php
@@ -1,6 +1,6 @@
 <?php namespace App\ApiResources;
 
-use App\Models\Bio as EloquentBio;
+use App\Bio as EloquentBio;
 use Illuminate\Contracts\Support\Arrayable;
 
 class Bio implements Arrayable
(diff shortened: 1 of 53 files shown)


mattstauffer:
@Xethron This is great, but I'd prefer they're under the `App` namespace instead of `App/Model`. Mind updating for that? Thanks!

Xethron:
Sure, no problem. However, I'd like to disagree with the request first :P 

I decided to add them to `App\Models` because there are already "non-models" in the /app directory (macros.php and modelEvents.php). Having them in their own namespace (folder) makes it easier to find all the models. Also searching in an IDE, typing App\Models\... will allow you to easily filter the list, where App\ will list everything in the App namespace

Lemme know what you think, else I'll update them later tonight.

mattstauffer:
@Xethron Yah, I understand the reasoning. But I prefer avoiding "purpose-specific" namespaces as much as possible (/Models, /Repositories, etc.), as much as they might convenience the IDE.

`macros.php` and `modelEvents.php` are definitely candidates to be moved to another folder--I'm open to suggestions!! :)

Thanks!

Xethron:
Awesome, no worries. Will fix that up as soon as I get a moment.

Xethron:
Nevermind, I forgot how easy it is in PhpStorm to change Namespaces, especially now that everything is using the ::class string resolution. :D

Xethron:
Hold up, don't merge...

Xethron:
Oh phew, was only the test factories! Ok, Tests are green!

Xethron:
Oh no...

Xethron:
Ok, I rebased the changes of master. Should I just overwrite the commits, or do you want me to open a new PR?

Xethron:
Created new PR #196 with the rebased changes. Closing this one.

Issue #127, opened 2015-09-16
Improve public pages


rdohms:
Practically the whole website is locked behind login, maybe we could start opening up a few pages? Off the top of my head the conference page is a good one to open up.

With the profile stuff coming along then maybe we should give the front page more life with some stats and lists?

Could also make the front page more like the pronto page i cooked up, list some stats, show the benefits, like being able to use pronto and etc...

mattstauffer:
@rdohms BTW, I couldn't agree with this more. <3

Xethron:
Ok, so Conferences are now open. As for stats, you thinking like number of:
 - Speakers Registered
 - Talks created
 - Talk Reversions
 - Conferences
 - Submissions to conferences

That kinda stats?

Perhaps also: Average number of talk submissions per speaker to conferences, average number of active talks per speaker... Those could be interesting.

(Ok, I've lost it again) And maybe, list the speaker with the most public talks, as well as the talk with the most reversions :D

I'll stop there :)

mattstauffer:
Closing until we get more concrete proposals for public pages. Thanks for suggestions and PRs!

PR #192, opened 2016-10-25
#127 Make conference index page public

Xethron:
This PR includes the contents of #191 as to avoid future merge conflicts. If #191 is declined, this one should also be declined


mattstauffer:
:+1:Thanks!

Issue #198, opened 2016-11-02
CSS for buttons on the show talk page messes up the flow with long titles


mattstauffer:
![image](https://cloud.githubusercontent.com/assets/151829/19931375/a246ffb0-a0e2-11e6-907f-e64c1fa84f2a.png)

Xethron:
Ok, changing `.page-title` to display block will change the layout on some pages, and there are only 3 places where user input is allowed in a `.page-title`. Two for talk title, and one for bio->nickname.

Option 1: create a class `.page-title-block` and use it in these three places
Option 2: create a class `.block` and override the display to block
Option 3: create a class `.page-title-inline` and change it everywhere else

I like Option 1...

mattstauffer:
I'd prefer a global solution that makes this work the way that's expected without having to differntiate based on user input or not. I feel like if i have some time to throw my CSS brain at it I'l find it but just haven't found that time yet.

PR #199, opened 2016-11-21
Fix problems with long talk titles

Xethron:
So I decided to look at #198 again to find an elegant way to fix this. I had a look at all views using .page-title, and in every place it was followed by a block element, except in the talk pages index and archived, which was followed by an a tag disguised as a button.

I have moved the button next to the Sort options, and I believe it actually looks a lot better here:
image

Let me know what you think :)


mattstauffer:
I don't *love* the button there but I do love a simple fix and maybe we can make the button make more sense somewhere else later. Thanks for this!!

Issue #201, opened 2016-12-29
Needs better handling of active/archived talks


mattstauffer:
- I think it'd be better for talks to global scope to ignoring archived talks, but you can add them back with a scope modifier, like soft deletes
- We should make it possible to request a list of talks including archived talks, or just archived talks, in the API
- ?

andrewmile:
![](https://github.trello.services/images/mini-trello-icon.png) [Globally ignore archived talks](https://trello.com/c/gkBhVzJV/92-globally-ignore-archived-talks)

PR #202, opened 2016-12-29
Drop Archived talks from API calls

mattstauffer:


mattstauffer:
Reference #179 ... future work: #201

PR #206, opened 2017-01-19
updated encryption type

timacdonald:
This should fix the issue raised here on twitter: https://twitter.com/vmitchell85/status/822105032107626498

Regarding the out-dated encryption method. I quick search, but couldn't find any encryption in the project, but please confirm before merging!

Thanks,
Tim

### DIFF
--- a/config/app.php
+++ b/config/app.php
@@ -83,7 +83,7 @@ return [
 
     'key' => env('APP_KEY', 'YourSecretKey!!!'),
 
-    'cipher' => MCRYPT_RIJNDAEL_128,
+    'cipher' => 'AES-256-CBC',
 
     /*
     |--------------------------------------------------------------------------


timacdonald:
this is actually still being discussed and may not be the appropriate solution

timacdonald:
If this is suitable, and encryption isn't actually used in the app,  would you like me to update the readme:

---

# Requirements

- PHP >= 5.5.9
- ~~PHP mcrypt extension~~


---

and composer.json

```
        "lucadegasperi/oauth2-server-laravel": "5.1.4",
        "ext-mcrypt": "*", // ditch ??
        "recaptcha/php5": "^2.0",
```

vmitchell85:
Looks like changing the encryption method does in fact allow everything to run fine. The McryptEncrypter class isn't used with the change in method so the deprecated function isn't ran.

mattstauffer:
This was already handled as a part of the Laravel Shift upgrade--but thanks @vmitchell85! :)

PR #208, opened 2017-01-20
Hide underline for public profile pic link

vmitchell85:
The underline was showing a single character after the image on hover. This removes the underline. See Image:

image

P.S. - I'm a little OCD...


mattstauffer:
haha thanks @vmitchell85!

PR #209, opened 2017-01-21
If applied this commit updates the asset's url's

introwit:
**The problem**
-  When cloned and ran locally , the assets such as the CSS , JS files , few routes & img's were not being loaded as the src url generated for them was not pointing to the correct location.

**Solution**
- Asset's src url's updated to use laravel's url() helper to generate
fully qualified correct url to the asset location.


mattstauffer:
@introwit Thanks so much!

PR #210, opened 2017-01-24
If applied this commit fixes the incorrect position of sign-up & login error messages on the home page

introwit:
**The Problem :** 

 - When you enter wrong credentials while logging in , the error messages get displayed in the SIGN-UP box rather than the login box as shown in the screenshot below.

fireshot capture 029 - symposium for speakers conferences - https___symposiumapp com_

- This was happening because the sign-up form partial had the error handling code block at it's top but the login form didn't. And if we added the error handling block to that as well then as expected same errors were displayed on both the places.

**Steps to recreate :**

- Try to login in the live copy with wrong credentials

**The Solution :**

- As the Laravel documentation itself suggests that when you have 2 or more forms on the same page , you should name the error bags for better & correct error handling. Docs : section 3 of 'Manually creating validators' with the title "Named Error Bags" : https://laravel.com/docs/5.3/validation#manually-creating-validators

- This commit overrides the sendFailedLoginResponse() method of the AuthenticateUsers trait in the LoginController to give the error bag a name.


mattstauffer:
Thanks @introwit!

mattstauffer:
For posterity, turns out there's a faster way to do this:

https://github.com/tightenco/symposium/commit/f7422d8d2cebf1d35313ee1d8235297aaae0f790

introwit:
@mattstauffer seems like I missed that. Thanks , learned a new thing :) but just out of curiosity may I know is there any specific reason behind that reordering of the imported namespaces in that commit? Would help me learn. Sorry for the silly question though!

mattstauffer:
@introwit Sure! I prefer to sort the import block by alphabetical instead of by statement length. :)

introwit:
Oh okay! Thank you Matt :) Will keep this in mind for the future contributions.

PR #213, opened 2017-02-16
Allow blacklisting email domains for signup

mattstauffer:


mattstauffer:
Needs to handle subdomains before merging.

besologic:
Subdomains handled!

Issue #166, opened 2016-04-19
[Proposal] Ability to search public speaker profiles


greydnls:
As a conference/UG organizer it would probably be helpful to filter speaker profiles by location. 

I think this would require adding in some additional fields to speaker's bio for airport or locale, but that could be done easily enough.

mattstauffer:
:+1:

Xethron:
I think this should be on the users profile instead of their bio... And perhaps even a "traveling" option, that allows you to set a from/to date in a different location that you will be open for presenting...

So you have a "Default", and a travel "Open for speaking". That way, when you search, you'll also find speakers with travel times close to that location in the future. (Phase 2)

mattstauffer:
source for some airport data for when we add this:

https://github.com/opencfp/opencfp/blob/2f747fc219b73f9b0a11308083d2a356056752a4/migrations/20151214140200_airports.php

mattstauffer:
(closed because we added it)

PR #214, opened 2017-02-21
Adds ability to search public speaker names and locations

josecanhelp:
References #166 

I utilized the [Google Places API Web Service](https://developers.google.com/places/web-service/intro) to submit user location on the account page. This was done to standardize the location input and improve speaker searches.

I also utilized [Laravel Scout](https://laravel.com/docs/5.3/scout) with [Algolia](https://www.algolia.com/) for full text searches. Currently, I have set it up to search specifically for user names and locations.

The following environmental variables will need to be added:
- GOOGLE_MAPS_API_KEY
- ALGOLIA_APP_ID
- ALGOLIA_SECRET

If you want to index an already populated application database, please run:
`php artisan scout:import "App\User"`


app/Http/Controllers/PublicProfileController.php
```diff
@@ -29,6 +29,23 @@ public function index()
             ->with('speakers', $users);
     }
 
+    public function search(Request $request)
+    {
+        $users = User::search($request->search_query)
```
mattstauffer: Let's rename this `query` and access it using `$request->get('query')` just for consistency.

app/Http/Controllers/PublicProfileController.php
```diff
@@ -29,6 +29,23 @@ public function index()
...
+            ->orderBy('name', 'asc')->get();
+
+        // Since Scout searches can only perform rudimentary where clauses,
+        // we must filter search results to only validly public profiles.
+        $filteredUsers = $users->filter(function ($user) {
```
mattstauffer: Huh. So you can't do

```php
$users = User::search($request->input('query'))
    ->where('enable_profile', true)
    ->where('profile_slug', "!=', NULL);
```

.. or something like that? I don't know, but that seems like almost as simple of a where clause as you can get, right?

It's also worth considering whether we only index public profiles, instead of users, which means we'll handle this consideration on the indexing side. Not sure how easy that is.

app/User.php
```diff
@@ -6,10 +6,11 @@
 use Illuminate\Notifications\Notifiable;
 use Illuminate\Support\Facades\Storage;
 use Thomaswelton\LaravelGravatar\Facades\Gravatar;
+use Laravel\Scout\Searchable;
```
mattstauffer: Please alphabetize use statements

config/scout.php
```diff
@@ -0,0 +1,62 @@
...
+    |--------------------------------------------------------------------------
+    |
+    | Here you may configure your Algolia settings. Algolia is a cloud hosted
+    | search engine which works great with Scout out of the box. Just plug
+    | in your application ID and admin API key to get started searching.
+    |
+    */
+
+    'algolia' => [
+        'id' => env('ALGOLIA_APP_ID'),
```
mattstauffer: Let's add these two to the `.env.example` file

config/services.php
```diff
@@ -35,4 +35,10 @@
         'secret' => env('STRIPE_SECRET'),
     ],
 
+    'google' => [
+        'maps' => [
+            'key' => env('GOOGLE_MAPS_API_KEY')
```
mattstauffer: Let's add this to the `.env.example` file

mattstauffer: Also, this is complex enough to get (if I remember correctly) that we should add instructions to the README for how to get them

resources/views/account/edit.blade.php
```diff
@@ -96,3 +110,53 @@
         {{ Form::close() }}
     </div>
 @stop
+
+@push('scripts')
+<script type="text/javascript">
+var autocomplete;
+var componentForm = {
+    neighborhood: 'long_name', // city within borough (Williamsburg)
```
mattstauffer: Is this "neighborhood within borough" not "city within borough"?  I don't know if this is a typo in your comment or if it's just confusing :)

resources/views/account/edit.blade.php
```diff
@@ -96,3 +110,53 @@
         {{ Form::close() }}
     </div>
 @stop
+
+@push('scripts')
+<script type="text/javascript">
```
mattstauffer: This is complex enough that I think I'd like to pull it out to a separate JS file that goes through Elixir and then gets included here.

app/Http/Controllers/PublicProfileController.php
josecanhelp: Per the Scout documentation:

> Currently, these clauses only support basic numeric equality checks, and are primarily useful for scoping search queries by a tenant ID.

Reading through Scout documentation and code, there is no way that I see to index a profile since the profile is a view and not a model. I will think about this one a bit more to see if there's an option I haven't thought of yet. For now, I think filtering the results is our best option.

config/scout.php
josecanhelp: These environment keys have been added to the .env.example file

config/services.php
josecanhelp: I created a separate guide so that we didn't clutter the main README. There is a link to it under `Requirements`.

josecanhelp:
Thanks for the great feedback @mattstauffer.

We are using the location. A site user can go to the `speakers` page and use the search bar that I placed there. 

Since the Algolia user index is storing name and location data, it will search both of those parameters to return a collection of users that match the criteria.

Screencasts below:

### Setting your location
set speaker location

### Searching speakers
search speakers

app/Http/Controllers/PublicProfileController.php
mattstauffer: Got it. 👍🏻 

And yah, what I meant by indexing profiles, not users, is just a mental model sort of thing; that is, indexing *for the sake of showing the profile*, which means we could totally exclude indexing non-public-profile people. But again, I can't even remember if Scout can do that. So I'm sure this is fine for now.

config/services.php
mattstauffer: :+1:

google-guide.md
```diff
@@ -0,0 +1,22 @@
+## Google API Key Guide
+
+### Create a Key
+
+1. Create an API Key from the [Google Places API Documentation](https://developers.google.com/places/web-service/get-api-key) by clicking on the "GET A KEY" button.
+
+![Generate API Key](https://www.dropbox.com/s/08jv8ngi8nm9r6x/google_key.gif?raw=1)
```
mattstauffer: Your list breaks here. I think you may need to indent this line by four spaces?

resources/views/account/edit.blade.php
```diff
@@ -85,6 +85,20 @@
...
                 </div>
+
+                <div class="form-group">
+                    {{ Form::label('location', 'Location', ['class' => 'control-label']) }}
+                    <span class="help-block">Enter the city in which you reside and local conference organizers can find you.</span>
+                    {{ Form::text('location', null, ['class' => 'form-control', 'id' => 'autocomplete']) }}
+                </div>
+
+                <div class="form-group">
+                    {{ Form::hidden('neighborhood', null, ['id' => 'neighborhood', 'readonly' => true]) }}
```
mattstauffer: I guess we should set the values of these to pull the old values; something like

```html
{{ Form::hidden('neighborhood', old('neighborhood, $form->neighborhood), ['id' = 'neighborhood', 'readonly' => true]) }}
```

mattstauffer: (This is necessary since I had you drop the conditionals

josecanhelp: This is done automatically when using `Form::model` according to [Laravel Collective documentation](https://laravelcollective.com/docs/5.3/html#form-model-binding). 

I tested this out myself and it works just fine.

google-guide.md
josecanhelp: Good catch, I missed that. Thanks.

resources/views/account/edit.blade.php
mattstauffer: Ah, right. Always forget about that. 👍🏻

imjohnbon:
@mattstauffer Just making a note that I reviewed this PR and didn't see any issues.

mattstauffer:
@josecanhelp Thanks so much for this! Very excited to see what all we can do with this. @imjohnbon Thanks for reviewing it!

PR #215, opened 2017-05-16
Replace favoriting with dismissing

imjohnbon:
See this issue for more information: https://github.com/tightenco/symposium/issues/66

Essentially, the concept of "favoriting" a conference was replaced with the ability to "dismiss" a conference. Dismissing a conference removes it from the users conference listing page.

A few notes:

1. I added a "dismissed" filter link in the filters group, just in case a user accidentally dismissed a conference that they didn't mean to. I'm not sure if this is ideal, it was just the simplest implementation. Another idea I played around with was having a section in your account dashboard that listed out the dismissed conferences.
2. By default, "All Time" does not include dismissed conferences. I was a bit torn on this, since "All Time" kind of implies _everything_, but I fell back to the idea that a user would not want to see dismissed conferences under any circumstances.
3. I used an "x" and a "+" icon for dismissing/undismissing respectively. I think they work alright, but I'm not sure they convey what you're doing as well as the star did for favoriting.
4. As part of this effort, I wrote a migration that will drop the "favorites" table. Which naturally would cause us to lose any data that's in there now. Nothing to really note here, just wanted to mention it as dropping tables isn't something I do lightly.


imjohnbon:
@mattstauffer I'm going to leave this for you to review since I wrote it and probably shouldn't review my own work 😄

mattstauffer:
Looks great! Only issue I'm seeing is that I'm surprised by this UI:

image

Is my Gulp building right? Or is that how you intended it? If that's how you intended it, we'll need to work a little bit of design into that dismiss button.

imjohnbon:
That's definitely not how I remember it looking when I worked on it, let me recompile this branch locally and see if I can replicate.

imjohnbon:
@mattstauffer I had to do a few hard-refreshes after pulling this down, but this is how it looks for me (and how I intended it to look):

screen shot 2017-06-21 at 11 14 10 am

Can you take another look when you have a chance and let me know if you still can't get it to look right?

mattstauffer:
Also--we need to hide the dismiss button when someone's not logged in.

imjohnbon:
@mattstauffer I like that actually, since everyone who uses the site is going to immediately do the same thing mentally (either write it off and ignore it, or think "I need remember this one"). So having the UI follow that mental model could be an interesting approach.

Regarding "we need to hide the dismiss button when someone's not logged in", I followed [the same convention we use right now for favoriting](https://symposiumapp.com/conferences). That being, the button shows up for anonymous users, but if they click on it it asks them to login. Of course we can divert from this now, but just wanted to bring that up before I hide it completely.

andrewmile:
Closing this PR as PR #248 as added the dismiss conferences feature along side the existing favoriting feature.

PR #219, opened 2017-06-12
Add route for Show Active Talks link

mrmercan:


mattstauffer:
Thanks for the PR! Just one question: what made you decide to explicitly pass the alpha sort, instead of letting it fall back to its default?

Issue #218, opened 2017-06-12
Bios not storing as public


michaelrtm:
A line similar to this `$bio->public = $request->get('public') == 'yes';` is missing from the store function on Bios, meaning by default all Bios are stored as private. It is in the update method.
Same goes for talks.

michaelrtm:
Was incorrect about talks

PR #220, opened 2017-06-13
Bio marked as public will save as such

michaelrtm:
This is the fix for #218 
Turns out I was incorrect about the talks being published correctly, so solves the issue.


mattstauffer:
@adamwathan In your idea about the form object, where would you put sanitizing/formatting input? Curious about how you normally handle this.

@michaelrtm Would you mind updating this branch where the commit message is more clear w/r/t what it's doing? Thanks!

michaelrtm:
@mattstauffer updated commit message :)

mattstauffer:
@michaelrtm Thank you!!

Issue #203, opened 2017-01-19
Add a calendar


mattstauffer:
For speakers who spend a lot of their year speaking, it'd be nice to have a calendar of important dates.

As a side idea, we may want to track information that's a little more flexible--when do I need to book my flights by? Reminder to get reimbursed! Almost like an arbitrary calendaring system.. but careful that we don't want to get too far down that road.. not exactly sure how we should do that TBH. 

But we can at least start with a calendar of known events. Probably "CFP close/open for conferences I have favorited" and "conference start/end dates for conferences I have favorited" and "conference start/end dates for conferences I'm accepted to speak at."

imjohnbon:
I started working on this a bit today. Still needs a lot of style cleanup, but I do have a calendar with *all* Conferences being output to it:

<img width="1008" alt="screen shot 2017-06-20 at 6 35 41 pm" src="https://user-images.githubusercontent.com/1489975/27358994-4c13f070-55e7-11e7-9b41-c19d4b895a59.png">

Currently just trying to figure out what we might want to filter the calendar down to. "Favorited" conferences was mentioned in the original message here, but it's possible that "favorites" could be going away per [this PR](https://github.com/tightenco/symposium/pull/215).

If that PR does get merged, I'm not exactly sure what else we could filter by. Except maybe "conferences I'm accepted to speak at", but I'm not sure where the "accepted talks" functionality stands either. I could work on that too if need be.

mattstauffer:
@imjohnbon I think you're right--"all conferences I have not dismissed" is a good start.

Then from there, probably "conferences I'm accepted to speak at."

We could also put CFPs in there in a different color.

PR #225, opened 2017-06-22
Add Calendar

imjohnbon:
See #203 for reference.

Major effort here was adding a /calendar page that lists out Conference dates, as well as CFP dates. A few notes:

1. I had to move jQuery up to the header for full calendar to work. Not ideal, but it wouldn't render without it.
2. Right now, the calendar lists out all events. When the "[replace favoriting with dismissing](https://github.com/tightenco/symposium/pull/215)" PR is merged, we can update this so it only pulls un-dismissed conferences.

Here's a screenshot of the calendar filled with events. CFP dates are gold, conferences are blue:

screen shot 2017-06-22 at 11 20 45 am


app/Http/Controllers/CalendarController.php
```diff
@@ -0,0 +1,47 @@
...
+
+use App\Conference;
+use Calendar;
+
+class CalendarController extends BaseController
+{
+    public function index()
+    {
+        $conferences = Conference::all()->map(function ($conference) {
+            return \Calendar::event(
```
mattstauffer: Don't need `\` if you imported it

app/Http/Controllers/CalendarController.php
```diff
@@ -0,0 +1,47 @@
...
+namespace App\Http\Controllers;
+
+use App\Conference;
+use Calendar;
+
+class CalendarController extends BaseController
+{
+    public function index()
+    {
+        $conferences = Conference::all()->map(function ($conference) {
```
mattstauffer: at least to start, let's make each of these calls a private method. We may also want to extract to another class or onto the conference class, not sure.

app/Http/Controllers/CalendarController.php
```diff
@@ -0,0 +1,47 @@
...
+                [
+                    'color' => '#428bca',
+                    'url' => route('conferences.show', $conference->id),
+                ]
+            );
+        })->toArray();
+
+        $cfps = Conference::all()->map(function ($conference) {
+            return \Calendar::event(
+                'CFP\'s open for ' . $conference->title,
```
mattstauffer:
```php
"CFPs open for {$conference->title}",
```

resources/views/calendar.blade.php
```diff
@@ -0,0 +1,22 @@
+@extends('layout')
+
+@section('headerScripts')
```
mattstauffer: I think you could move moment, fullcalendar, jquery, AND the inlin jquery call all into the footer and make it work. hit me up if you have questions how

app/Http/Controllers/CalendarController.php
imjohnbon: Yeah, I generally try to avoid private controller methods like the plague, but it doesn't feel quite right on the Conference class either. I think private methods are good for now. Maybe as the calendar functionality grows later on, the proper pattern will become more clear.

imjohnbon:
@mattstauffer Thanks for all the input. I was able to move jQuery and the rest of the jQuery-related stuff back into the footer. Take a look at my implementation and let me know if it's what you had in mind.

app/Http/Controllers/CalendarController.php
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    private function cfpEvents()
+    {
+        return Conference::all()->map(function ($conference) {
+            return Calendar::event(
+                "CFPs open for {$conference->title}",
+                true,
+                $conference->cfp_starts_at,
+                $conference->cfp_starts_at,
```
mattstauffer: should this not be the ends at?

resources/views/calendar.blade.php
```diff
@@ -0,0 +1,21 @@
...
+    <div class="container body">
+        <div class="row">
+            <div id="calendar" class="calendar col-sm-12">
+                {{ $calendar->calendar() }}
+            </div>
+        </div>
+    </div>
+@stop
+
+@push('scripts')
```
mattstauffer: 👌

imjohnbon:
@mattstauffer Good catch there. I think it actually brought up an interesting UI problem too. Because CFPs can be open for so long, it adds huge streaks of orange to the calendar:

screen shot 2017-06-23 at 1 13 31 pm

Do you see this as an issue? As a potential fix, we could add 2 single day events for CFP's (one event on the start date, and another event on the end date).

mattstauffer:
@imjohnbon ugh that's tough! I prefer the bars, but you're right, it's gonna be too much. Single day events it is.

imjohnbon:
@mattstauffer Cool! I split up CFP's into 2 events. I made the "ending" event red in an effort to make things more clear, visually. Let me know if you're on board with the color choices.

screen shot 2017-07-12 at 4 27 52 pm

PR #211, opened 2017-01-25
If applied this commit implements a flexible social login mechanism with github login completely configured & working already

introwit:
**The social login mechanism that has been implemented tends to be completely flexible and easy to understand having the following pros:**

- Routes & controller methods are completely dynamic for any given social login service.
- A middleware to protect & check login attempts against our defined social login services.
- Any social login can be initiated within a single minute of configuration (How that can be done is explained later in this comment)
- Any social login can be revoked within a minute without affecting other logins.
- **Handles all the duplicate logins & signups scenario with breeze**

The Code flow : 

- Just 2 flexible routes to handle all of the social login services
- I have tried to keep the SocialLoginController code as readable as possible & _yes I have reordered the imported namespaces alphabetically as you taught me last time_ :)
- Added a middleware to protect & check login attempts against our defined social login services in the _config/social.php_

**Quick guide**

- For the "GitHub" sign in button the css has been added to a new _github.scss file and imported at the top in app.scss and that has been compiled already with Laravel elixir.
- As expected Laravel Socialite is being used so that should be installed using `composer install` 
_NOTE : SOCIALITE'S V 2.0 SHOULD BE USED WITH OUR 5.3 APP SO  a composer install command is recommended as the composer.json has v2.0 for it._
- Next to migrate the new required social login table run `php artisan migrate`
- Lastly update your .env with GitHub app credentials , for correct keys refer to config/services.php

**To Add a new social login**

- Just add the credentails in .env & update the config/services respectively.
- Update the config/social.php array to keep the middleware check active for it.
- That's it! 

**To Remove any login at any time without affecting other stuff**

- Just remove the respective one from config/social.php
- And you are done!

Sorry for the long read , I hope you don't mind Matt :)

Lastly , here is the screenshort of the github button implemented! Tried hard to make it look as near as possible with @adamwathan 's design in this github login issue #56  

github_ui

Sorry Adam if its not as good as you expected, I am not good with the front-end stuff but still tried my best :)

PR #226, opened 2017-07-15
If applied this commit implements a flexible social login mechanism with already active GitHub login

introwit:
> Better version of #211 

**Code Flow:**
- 2 flexible routes to handle all social login services.
- An array in `config/social.php` file that lists the services to be allowed to pass through the middleware.
- A middleware `app/Http/Middleware/Social.php` that checks is the incoming social login allowed in the above array or not.
- A migration `CreateUserSocialsTable` to create the new table that will consist of all the social login data.
- Covers all duplicate logins & signup scenarios.

**Steps to get this working:**

- Run `composer update` since Laravel Socialite is added.
- Run `php artisan migrate` to migrate the new required social login table.
- Update the `.env` file with GitHub credentials.

**To add & remove a new social login:**

- Update the array in `config/social.php` file to allow the respective social login to pass the middleware.
- Remove the name from the array in `config/social.php` file so the middleware will not allow it go through.

**UI**

---

PS: Ordered imported namespaces alphabetically as @mattstauffer taught me & CSS properties are also alphabetized as learned from @imjohnbon :relaxed:


app/Http/Controllers/Auth/SocialLoginController.php
```diff
@@ -0,0 +1,55 @@
...
+use App\Http\Requests;
+use App\User;
+use Illuminate\Http\Request;
+use Socialite;
+
+class SocialLoginController extends Controller
+{
+    public function __construct()
+    {
+        $this->middleware(['social', 'guest']);
```
imjohnbon: Let's move this middleware declaration out of the controller and into the routes file (in a route group or something if it effects multiple routes).

app/Http/Controllers/Auth/SocialLoginController.php
```diff
@@ -0,0 +1,55 @@
...
+        return Socialite::driver($service)->redirect();
+    }
+
+    public function callback($service, Request $request)
+    {
+        $serviceUser = Socialite::driver($service)->user();
+
+        $user = $this->getExistingUser($serviceUser, $service);
+
+        if (!$user) {
```
imjohnbon: Let's add a space after the !

app/Http/Controllers/Auth/SocialLoginController.php
```diff
@@ -0,0 +1,55 @@
...
+            $user = User::create([
+                'name' => $serviceUser->getName(),
+                'email' => $serviceUser->getEmail(),
+            ]);
+        }
+
+        if (!$user->hasSocialLinked($service)) {
```
imjohnbon: Let's add a space after the !

app/Http/Middleware/Social.php
```diff
@@ -0,0 +1,24 @@
...
+    /**
+     * Handle an incoming request.
+     *
+     * @param  \Illuminate\Http\Request  $request
+     * @param  \Closure  $next
+     * @return mixed
+     */
+    public function handle($request, Closure $next)
+    {
+        if (!in_array(strtolower($request->service), array_keys(config('social.services')))) {
```
imjohnbon: Let's add a space after the !

app/User.php
```diff
@@ -119,4 +119,13 @@ protected static function boot()
             \DB::table('favorites')->where('user_id', $user->id)->delete();
         });
     }
+
+    public function social()
+    {
+        return $this->hasMany(UserSocial::class);
+    }
```
imjohnbon: Let's add a newline after this method.

resources/assets/sass/app.scss
```diff
@@ -433,3 +433,141 @@ legend {
...
+
+.btn-auth:hover,
+.btn-auth:focus,
+.btn-auth:active {
+    color: #222;
+    text-decoration: none;
+}
+
+.btn-auth:before {
+    background: url(https://raw.githubusercontent.com/necolas/css3-social-signin-buttons/gh-pages/auth-icons.png) no-repeat 99px 99px;
```
imjohnbon: Let's move this background image to the repo instead of loading it from a third party URL.

introwit:
@imjohnbon fixed everything else, but running `gulp` seems to make the img url go back to the old one. Did some research and this seems to be a known issue. Why these automation tools are so confusing :cry: Let me know should I push the remaining changes or the next step to make. Will research and try to solve this myself till then.

imjohnbon:
@introwit Thanks for the update! You can feel free to push the remaining changes that you had worked on. If the image issue is still happening in Gulp, I can take a look at it later today :).

introwit:
@imjohnbon done :)

imjohnbon:
@introwit Thanks! I was able to make the update to load the background image locally. Not sure what was going wrong before, is it possible you were accidentally editing the compiled `app.css` instead of the `app.scss`? That's the only thing I can really think of. Anyway, this is looking good to me. I'm going to pass it along for final review. 😄

introwit:
@imjohnbon I was editing `app.scss` only but after running `gulp` it was going back to the previous url. I know it sounds weird but that was the thing happening 😅 Thank you for approving the changes, learned new things because of this PR ❤️

database/migrations/2017_07_15_125034_CreateUserSocialsTable.php
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    /**
+     * Reverse the migrations.
+     *
+     * @return void
+     */
+    public function down()
+    {
+        Schema::dropIfExists('user_socials');
```
josecanhelp: For the down migration, we need to 
1. drop the foreign key constraint
2. fix the misspelling of the table name

It should look like this:

```php
public function down()
    {
        Schema::table('users_social', function (Blueprint $table) {
            $table->dropForeign(['user_id']);
        });
        
        Schema::dropIfExists('users_social');
    }
```

PR #234, opened 2018-05-01
Fix Oauth Routing

rdohms:
During the installation of Passport to be the new Oauth/Api provider,
the Oauth routes were removed (https://github.com/tightenco/symposium/commit/62119f19e0035f5b76630621757d360fed99bff2#diff-7e3ce459dfcc113722bdf4667ceffc11L87)

However in what looks like a bad merge they were re-added during the
install of social login, but they now point to an missing
OAuthController (https://github.com/tightenco/symposium/commit/9b1ecb6186bb77cf2744e90c09d75e3e21edf617#diff-7e3ce459dfcc113722bdf4667ceffc11R91)

This commit removes these routes once more.


rdohms:
This should fix rdohms/pronto#45

PR #205, opened 2017-01-19
Laravel 5.3 Shift

mattstauffer:
This pull request includes the changes for upgrading to Laravel 5.3. Feel free to commit any additional changes to the `laravel-5.3-shift` branch.

**Before merging**, you should:

- Checkout the `laravel-5.3-shift` branch
- Review **all** pull request comments for additional changes
- Run `composer update` (if the pre-commands fail, add `--no-scripts`)
- Thoroughly test your application

If you would like more help with your upgrade, check out the new [human services from Laravel Shift](https://laravelshift.com/human-services).

PR #236, opened 2018-10-10
Remove note about mcrypt extension

BackEndTea:
The project hasn't used the mcrypt cipher since #205, so the note about it being needed may confuse users.


mattstauffer:
Thanks!

BackEndTea:
Thank you @mattstauffer

PR #237, opened 2018-10-22
Add github to login view

mpociot:
I noticed that the "Sign in with GitHub" option was missing from the "Login" page, which was really confusing.


mattstauffer:
Thanks Marcel!

PR #238, opened 2018-10-29
Add install instructions for Algolia keys

christophrumpel:
This PR updates the installation guidelines with information about the Algolia keys. These are needed in order to run PHPUnit tests.


mattstauffer:
Thanks!

Issue #16, opened 2014-12-22
Adjust date sorts on conferences view page to be actually useful


mattstauffer:
Probably default to sort by CFP ending soonest, then maybe also add CFP starting soonest? Maybe conf-related?

mattstauffer:
This issue was moved to tightenco/symposium#16

PR #240, opened 2019-01-11
Default to Next CFP closing date and add sort by Next CFP start data

josecanhelp:
39136856-2bd5f62e-46ea-11e8-8dca-26bf1788d227

Resolves Issue #16


app/Http/Controllers/ConferencesController.php
```diff
@@ -46,28 +46,23 @@ public function index(Request $request)
...
-                // Pass through
+                $conferences = $conferences->sortBy->title;
+                break;
+            case 'closing_next':
+                // pass through
             default:
+                // Forces closed CFPs to the end.
                 $conferences = $conferences->sortBy(function (Conference $model) {
-                    return strtolower($model->title);
+                    return isset($model->cfp_ends_at) ? $model->cfp_ends_at : Carbon::now()->addCentury();
```
mattstauffer: So, the original logic did two things:

1. Show those which close in the future, sorted by how soon
2. Show those which have no CFP close date
3. Show those which have already closed.

This current re-work of it gets rid of #3 (in the order) and now, if a CFP closed 4 weeks ago, it'll be higher in the order than one which closes in a week, which defeats the purpose, even if it is more logical.

What do you think about that?

andrewmile:
@mattstauffer I have modified the conference index page sort order such that conferences with past CFP end date will sort to the bottom, conferences with with a null CFP end date will sort in the middle, and conference with future CFP end date will sort to the top.  In addition, conferences with a null or past CFP end date will now sort by the conference start_at, within their respective grouping.

app/Http/Controllers/ConferencesController.php
```diff
@@ -47,28 +47,34 @@ public function index(Request $request)
 
         switch ($request->input('sort')) {
             case 'date':
+                $conferences = $conferences->sortBy->starts_at;
+                break;
+            case 'opening_next':
```
mattstauffer: Would this sort a cfp start date in the past at the top? If so, I think we need to make the same modification here

andrewmile: I modified this so when sorting by date, past conferences will now sort below future conferences.

andrewmile:
@mattstauffer This is ready for your review after making your recommended changes.

app/Http/Controllers/ConferencesController.php
```diff
@@ -47,28 +47,40 @@ public function index(Request $request)
...
-                            return $model->cfp_ends_at;
-                        } elseif ($model->cfp_ends_at === null) {
-                            return Carbon::now()->addYear(500);
-                        } else {
-                            return $model->cfp_ends_at->addYear(1000);
-                        }
-                    });
+            case 'opening_next':
+                // Force CFPs with no CFP start date to the end
+                $conferences = $conferences->sortBy(function ($conference) {
```
mattstauffer: My comment about pushing to the end was about this one (opening_next), not the one you made the change on (date)

andrewmile: I just removed the modification to the date sort and tested to ensure the opening_next sort sorts null cfp_ends_at below non-null cfp_ends_at.  There were no changes needed to be made to the switch case to get the functionality we wanted.

andrewmile:
@mattstauffer I just updated this branch to ensure sorting by date and closing_next works as expected.

app/Http/Controllers/ConferencesController.php
```diff
@@ -47,28 +47,34 @@ public function index(Request $request)
...
-                // Pass through
+                $conferences = $conferences->sortBy->title;
+                break;
+            case 'closing_next':
+                // pass through
             default:
-                $conferences = $conferences->sortBy(function (Conference $model) {
-                    return strtolower($model->title);
+                // Forces closed CFPs to the end.
+                $conferences = $conferences->sortBy(function ($conference) {
```
mattstauffer: I think we said we were going to simplify this one down too, right?

andrewmile: Yep, thanks for catching that!  The default sort now sorts by `cfp_ends_at`, whether it is in the past or future.  Conferences with a `null` `cfp_ends_at` will be sorted to the end.

andrewmile:
@mattstauffer I simplified the default sort and this is ready for your review.

PR #239, opened 2018-10-31
Feature - Send Open CFP Notification

christophrumpel:
screen shot 2018-10-31 at 1 10 50 pm
This PR is a `first draft` for sending out email notifications for open CFPs. It's the first time I'm working on this project, so let me know if I did get something wrong.

**The PR includes:**

* Firing a new `event` for newly user-created conferences
* Adding a `listener` to trigger the notification if the newly created conference has an open CFP
* Adding the notification
* `Console Command` for checking conference CFPs (empty right now until we move on with this PR)
* `Tests` for notifications
* Resetting mail config so we can use the `.env` file custom setup. (Mailtrap in my case)

**Before I move on, we need to talk about a few things:**

1. `User trigger email`: Since all user can create conferences, they can trigger sending out emails. This could be a spam problem. 
2. `Joindin Importer`: Next to the user created conferences, there are the ones coming in from the importer. If you like this PR, it would be a good idea to fire the event there as well and send out emails there as well.
3. `Cronjob`: Next to checking open CFPs on a `create-conf-base` we also have to check them with a cronjob daily. This is why there is this empty console command already. Here we need to think about how we know if a notification was already sent, so we don't send out duplicates. Maybe a new field in the conference table?
4. `Opt-in`: We should add an option to opt-in to those email notifications from the user's account settings.  Best would also be to select categories, but I think this is something for another PR.
5. `Mail design/content`: This is just the basic Laravel Notification mail with a button to the `cfp_url`. Since this field is not required we could also put the conf URL in here as a default. Any other wishes for the content/design?

PR #241, opened 2019-01-20
Feature - Send Open CFP Notification (#239, with matt's modifications)

mattstauffer:


christophrumpel:
Hey, just looked through your commits. Looks good to me! So with Slack, you provide just the link again and then the admin checks the conf on the site and can approve it right? Maybe this is even better than approving right through the Slack message. This way an admin actually has to check the conference :-) Looking forward to trying these changes out 👍

mattstauffer:
@christophrumpel yep! :)

mattstauffer:
@christophrumpel We've got this up and running on staging.symposiumapp.com. Please feel free to test it out!

christophrumpel:
Niceee 👍

mattstauffer:
image

Just got my first one from the staging site. 11pm is a bit weird, so I'm guessing we have a server time zone issue.

christophrumpel:
Hmm the cronjob is running every minute right? Do you see the time difference between the time the conference was imported and the notification?

mattstauffer:
I guess it *could've*, but here's the record:

image

Feels super likely that 9am server time is 11pm my time... somehow. So I just need to dive in at some point and figure out the most appropriate space to set that sort of config.

image

The app is configured for Detroit, which is my time zone, so I'll have to do a deeper dive to try to figure out what's going on there.

Actually.. wait.. how did it even run?

image

It's set to only notify on prod?

I'm so confused. I'll have to take a closer look later.

christophrumpel:
Hey Matt, if I remember correctly it works like that:

- when a new conference is imported, it checks if it already accepts CFPs, if yes we automatically send out the notification
- the cronjob is only for cases where a conference was added where the CFP was not open yet

This would explain why you got the email almost the same time the conference was added. Right?

mattstauffer:
@christophrumpel oh my goodness I misread the created_at date. You're 100% right and i"m an idiot.

Thank you!

christophrumpel:
np at all. It is difficult to remember those things. I wrote it and even I wasn't sure about it :-)

Issue #204, opened 2017-01-19
Add speaker-specific metadata about conferences


mattstauffer:
https://twitter.com/TheLarkInn/status/821945642503307264

It'd be nice to be able to allow speakers to track more information about conferences they're speaking at all together in one place.

greydnls:
@mattstauffer do you think adding the concept of `Acceptances` to abstracts would allow speakers to track (and optionally expose) where an abstract has been accepted. Assuming the conference is also on the platform they could link it, and then easily add notes about travel, etc.

greydnls:
Mind if I take a stab at it?

mattstauffer:
I wouldn't mind at all!

andrewmile:
Closed by #413

PR #243, opened 2019-02-05
Add acceptances and implement UI for conference page

greydnls:
This PR: 
- [x] Addresses (in part) #204 
- [x] Elevates `Submission` to a first class citizen and adds primary key as well as unique index to the table.
- [x] Creates `acceptances` table and relates to `TalkRevision` and `Conference` and `Submission`. 
- [x] Adds UI to the Conference view page to indicate which talks have been submitted vs accepted as well as functionality to move from submitted to accepted.  

#### Open questions: 
1. I would like to add some UI on the Talk page to indicate which conferences a given talk has been submitted/accepted to, but I think that would be better suited for a separate PR, in the spirit of keeping small, manageable changes. 
3. The acceptance table at this point is just a pivot table with timestamps but the intention is to be able to add additional fields for meta data. Is there an initial set of meta data that you think would be useful?

#### Notes:
PR is a proof of concept and doesn't include tests. Once I get a 👍 on general approach, I'll add tests. 

#### Screen Shots: 
Accepted State:
screen shot 2019-02-04 at 11 29 33 pm

Submitted State:
screen shot 2019-02-04 at 11 50 45 pm

Unsubmitted State:
screen shot 2019-02-04 at 11 29 55 pm


app/Acceptance.php
```diff
@@ -0,0 +1,27 @@
...
+class Acceptance extends UuidBase
+{
+    protected $table = 'acceptances';
+
+    protected $guarded = [
+        'id'
+    ];
+
+    protected $fillable = [
+      'talk_revision_id',
```
mattstauffer: Indentation issues here

app/Commands/ConvertToAcceptance.php
```diff
@@ -0,0 +1,33 @@
+<?php
+
+namespace App\Commands;
+
+use App\Conference;
+use App\Talk;
+
+class ConvertToAcceptance extends Command
```
mattstauffer: Disclaimer: I haven't worked with Symposium in a while, so I may just be mis-remembering the architecture.

My gut on this said I'd rename it to something like "ConvertSubmissionToAcceptance."

But when I started typing that, I started wondering, why are we converting at all? It's not not-submitted once it's accepted. So what makes more sense to my brain is one of the following:

1. Add metadata to a submission. Maybe an `is_accepted` column.
OR
2. Don't convert a submission to an acceptance, but rather *add* an acceptance, so that a submitted talk is always submitted, and sometime's it's *also* accepted.

app/Http/Controllers/SubmissionsController.php
```diff
@@ -20,6 +21,18 @@ public function store(Request $request)
         return response()->json(['status' => 'success', 'message' => 'Talk Submitted']);
     }
 
+    public function update(Request $request)
```
mattstauffer: What if this was instead AcceptancesController@create()?

app/Http/Controllers/ConferencesController.php
```diff
@@ -111,14 +112,28 @@ public function show($id)
...
+        $acceptances = $conference->myAcceptedTalks();
+
+        $talks = auth()->user()->talks->map(function ($talk) use ($submissions, $acceptances) {
+            /** @var Talk $currentTalk */
+            $currentTalk = $talk->current();
+            return [
+                'id' => $talk->id,
+                'title' => $currentTalk->title,
+                'url' => $currentTalk->getUrl(),
+                'submitted' => $submissions->search(function ($item) use ($talk) {
```
mattstauffer: This could be moved onto the model, yah?

`$talk->submittedToConference($conference)`

same for accepted:

`$talk->acceptedToConference($conference)`

then they would have the same logic you have here.

app/Conference.php
```diff
@@ -176,9 +176,23 @@ public function myTalks()
         });
     }
 
+    /**
+     * Return all talks from this user that were submitted to this conference
```
mattstauffer: Out of date doc block, fwiw

app/Http/Controllers/SubmissionsController.php
greydnls: I can do that. I was being a bit lazy here. Wanted to get the data model right first.

app/Http/Controllers/ConferencesController.php
greydnls: :+1: I think that's cleaner, yes.

greydnls:
@mattstauffer I think the issue with tracking acceptances only through an `accepted_at` field on a Submission is that it kind of defeats the purpose of the original issue this PR aims to solve, which is to allow speakers to track meta data (I'm guessing flight, hotel, travel days, etc) on accepted talks. 

I do like the idea of keeping both submission and acceptance, though. It would allow for better analytics regarding submission conversion for speakers. 

One issue I was having with the existing data model is that `Submission` isn't a first class citizen, it's really a pivot table with a `status` field and timestamps, it doesn't have a primary key. That makes it rather difficult (at least for me) to find a single record. I'd like to convert `Submission` to a FCC with primary key and optionally link it to an `Acceptance`, which for the time being is a very sparse table but will become more expansive as we decide what meta data it makes sense to track there. 

Does that make sense?

Also, do you have a staging or pre-prod environment for Symposium? I'm worried about a possible BC break on the Submissions table. As I mentioned, there's no primary key, but there's also no unique constraint on the `talk_revision_id`-`conference_id` key, so you can end up with duplicative records that shouldn't be there. I'd like to change that as well.

mattstauffer:
I think that sounds like a great idea.

No staging environment at the moment, but I'd be more than happy to set one up.

mattstauffer:
Moved this to point at the new develop branch, which will auto-deploy to staging.symposiumapp.com once I get it set up. Also, there's an IRC chat (not very populated) on freenode at #symposium if that would ever be helpful.

app/Http/Controllers/ConferencesController.php
```diff
@@ -111,14 +112,25 @@ public function show($id)
             return redirect('/');
         }
 
-        $talksAtConference = $conference->myTalks()->map(function ($talkRevision) {
-            return $talkRevision->talk->id;
+        $talks = auth()->user()->talks->map(function ($talk) use ($conference) {
```
greydnls: I feel like this should be in a presenter/transformer layer, but I couldn't find a similar convention for it. If there is one, please point me at it.

app/UuidBase.php
```diff
@@ -12,4 +13,17 @@ class UuidBase extends Eloquent
      * @var bool
      */
     public $incrementing = false;
+
+
+    /**
+     * Boot function from laravel.
+     */
+    protected static function boot()
```
greydnls: I don't know if this is necessary, but I wasn't able to get UUID ids to generate for submissions or acceptances without it. If there's another way, please let me know.

greydnls:
@mattstauffer I think this is ready for another review, when you have the time. I removed the `Commands` as I felt they were an extra level of indirection to perform actions that once you have Submission ids becomes very trivial. If you prefer that architecture I can add them back in.

app/Acceptance.php
```diff
@@ -0,0 +1,32 @@
...
+    protected $guarded = [
+        'id'
+    ];
+
+    protected $fillable = [
+        'talk_revision_id',
+        'conference_id',
+    ];
+
+    public static function createFromSubmission(Submission $submission)
```
mattstauffer: 👌

app/Http/Controllers/AcceptancesController.php
```diff
@@ -0,0 +1,21 @@
...
+
+class AcceptancesController extends Controller
+{
+    public function store(Request $request)
+    {
+        $submission = Submission::findOrFail($request->input('submissionId'));
+
+        $acceptance = Acceptance::createFromSubmission($submission);
+
+        $submission->recordAcceptance($acceptance);
```
mattstauffer: I wonder if this would be reasonable to do within `Acceptance::createFromSubmission`?

app/Http/Controllers/ConferencesController.php
mattstauffer: Yah.. we could use API resources, or just a manual transformer for this. Since this doesn't end up in the API, I think I'd be more likely to make a basic transformer class for it.. something like

```php
class TalkTransformer
{
    public function __construct($talk)

    public function transform()
        return array
}
```

... probably

app/Http/Controllers/AcceptancesController.php
```diff
@@ -0,0 +1,21 @@
...
+
+use App\Acceptance;
+use App\Submission;
+use Illuminate\Http\Request;
+
+class AcceptancesController extends Controller
+{
+    public function store(Request $request)
+    {
+        $submission = Submission::findOrFail($request->input('submissionId'));
```
mattstauffer: We should guard to make sure this user has permissions on this submission.

mattstauffer: Maybe just auth()->user()->submissions()->findOrFail()

app/Http/Controllers/SubmissionsController.php
```diff
@@ -2,32 +2,31 @@
...
-        $this->dispatch(new CreateSubmission(
-            $request->input('conferenceId'),
-            $talk->id
-        ));
-
-        return response()->json(['status' => 'success', 'message' => 'Talk Submitted']);
+        $talk = auth()->user()->talks()->findOrFail($request->input('talkId'))->current();
+        $conference = Conference::findOrFail($request->input('conferenceId'));
+
+        $submission = Submission::create([
```
mattstauffer: Can we do:

```php
$conference->submissions()->create(['talk_revision_id' => $talk->id]);
```

?

Also, if we're passing a talk_revision ID, is this not the wrong ID?

app/Http/Controllers/SubmissionsController.php
```diff
@@ -2,32 +2,31 @@
...
+            'conference_id' => $conference->id,
+            'talk_revision_id' => $talk->id
+        ]);
```
mattstauffer: New line after the `]);` please

app/Http/Controllers/SubmissionsController.php
```diff
@@ -2,32 +2,31 @@
...
-    public function destroy(Request $request)
+    public function destroy($id)
     {
-        $talk = auth()->user()->talks()->findOrFail($request->input('talkId'));
-
-        $this->dispatch(new DestroySubmission(
-            $request->input('conferenceId'),
-            $talk->id
-        ));
+        Submission::destroy($id);
```
mattstauffer: The original (using `auth()->user()->talks()`) guards that this user has access to this submission to delete it. This is a way we handled ACL before we had ACL actually coded in. :)

app/Talk.php
```diff
@@ -73,4 +73,18 @@ public function scopeActive($query)
     {
         return $query->where('is_archived', false);
     }
+
+    public function getSubmission(Conference $conference)
+    {
```
mattstauffer: I wonder if it would be cleaner/clearer to do:

```php
return Submission::where('conference_id', $conference->id)
    ->whereIn('talk_revision_id', $this->talkRevisions()->pluck('id'))
    ->first);
```

Not sure which is better tbh... my brain is a bit shot at the moment. I"ll just throw it out there and you can choose.

Same for getAcceptance below.

app/UuidBase.php
mattstauffer: In the current setup, this should go in `app/modelEvents.php`. I'm OK if we move it to here as long as we also cover the deleting event in talk, and then delete the modelEvents file.

database/migrations/2019_02_04_214119_create_acceptances_table.php
```diff
@@ -0,0 +1,49 @@
...
+            $table->string('conference_id', 36);
+            $table->foreign('conference_id')
+                ->references('id')
+                ->on('conferences')
+                ->onDelete('cascade');
+
+            $table->unique(['talk_revision_id', 'conference_id'], 'acceptances_talk_revision_conference_unique');
+
+            $table->timestamps();
+
```
mattstauffer: delete two spare lines here please

database/migrations/2019_02_04_214119_create_acceptances_table.php
```diff
@@ -0,0 +1,49 @@
...
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::create('acceptances', function (Blueprint $table) {
+            $table->uuid('id');
```
mattstauffer: Is there any benefit to considering attaching an acceptance just directly to a submission instead of to a conference and a talk revision? Again, braindead so I',m not sure.. just curious.

database/migrations/2019_02_05_083929_add_primary_key_to_submissions.php
```diff
@@ -0,0 +1,39 @@
...
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('submissions', function (Blueprint $table) {
+            $table->uuid('id');
+            $table->primary('id');
+        });
+
+        \App\Submission::all()->map(function($submission){
```
mattstauffer: Please import Submission, and space after `function` and space between closing paren and opening brace please

database/migrations/2019_02_05_083929_add_primary_key_to_submissions.php
```diff
@@ -0,0 +1,39 @@
...
+            $submission->id = \Ramsey\Uuid\Uuid::uuid4();
```
mattstauffer: Please import Uuid

database/migrations/2019_02_05_084300_add_acceptance_id_to_submissions.php
```diff
@@ -0,0 +1,38 @@
...
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('submissions', function (Blueprint $table) {
+            $table->string( 'acceptance_id', 36)->nullable();
```
mattstauffer: Please drop space before `'acceptance_id`

database/migrations/2019_02_05_084300_add_acceptance_id_to_submissions.php
```diff
@@ -0,0 +1,38 @@
...
+            $table->foreign('acceptance_id')
+                ->references('id')
+                ->on('acceptances')
+                ->onDelete('cascade');
```
mattstauffer: We don't want to delete the submission if the acceptance is deleted.

database/migrations/2019_02_05_084300_add_acceptance_id_to_submissions.php
```diff
@@ -0,0 +1,38 @@
...
+     */
+    public function down()
+    {
+        Schema::table('submissions', function (Blueprint $table) {
+            $table->dropForeign('submissions_acceptance_id_foreign');
+            $table->dropColumn('acceptance_id');
+        });
+    }
+}
+
```
mattstauffer: Drop extra line please

database/migrations/2019_02_05_113705_drop_status_column_from_submissions.php
```diff
@@ -0,0 +1,32 @@
...
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('submissions', function (Blueprint $table) {
+            $table->dropColumn( 'status');
```
mattstauffer: Drop leading space please

app/Talk.php
greydnls: I updated the function names to reflect that we're getting submissions for a specific user, which I don't believe your proposed query would do?

database/migrations/2019_02_04_214119_create_acceptances_table.php
greydnls: I left it denormalized for the time being until I'm certain how we'll be querying for acceptances and because it's easier to drop columns than it is to add and backfill them. It may turn out that we don't need those relations and can drop them.

app/Http/Controllers/AcceptancesController.php
greydnls: That relationship is... involved. 

`user -> talks[] -> revisions[] -> submission -> acceptance->findOrFail().`

I tried a different way of guarding instead. Very open to your opinion there as I'm sure there's a nicer way to do it.

greydnls:
@mattstauffer Once more with feeling? And now with TESTS!

app/Http/Controllers/AcceptancesController.php
mattstauffer: Geez, so that's a user has many through many through, which definitely doesn't exist natively.

I don't see an existing guard in here--is there one? I don't see anything that ensures that the currently-logged-in user has access to the submission.

Let's try it from the other angle: check that the `$submission` is owned by current user.

```php
$submission = Submission::findOrFail($request->input('submissionId'));

if (auth()->user()->id !== $submssion->talkRevision->talk->user_id) {
    abort(401);
}
```

maybe?

app/Http/Controllers/ConferencesController.php
```diff
@@ -111,14 +112,25 @@ public function show($id)
...
+            /** @var Talk $currentTalk */
+            $currentTalk = $talk->current();
+            $submission = $talk->getSubmission($conference);
+            $acceptance = $submission->acceptance;
+            return [
+                'id' => $talk->id,
+                'title' => $currentTalk->title,
+                'url' => $currentTalk->getUrl(),
+                'submitted' => $acceptance ? false : !!$submission,
```
mattstauffer: Hm, I think of it as submitted even if it's accepted--can we pass it that way and then update the template to still do what you intend?

app/Http/Controllers/SubmissionsController.php
```diff
@@ -2,32 +2,31 @@
...
     {
-        $talk = auth()->user()->talks()->findOrFail($request->input('talkId'));
-
-        $this->dispatch(new CreateSubmission(
-            $request->input('conferenceId'),
-            $talk->id
-        ));
-
-        return response()->json(['status' => 'success', 'message' => 'Talk Submitted']);
+        $talk = auth()->user()->talks()->findOrFail($request->input('talkId'))->current();
```
mattstauffer: Ahh--so since this is `current()`, which I totally didn't notice, can we name this variable `$talkRevision` so it's clearer?

app/Http/Controllers/SubmissionsController.php
```diff
@@ -2,32 +2,31 @@
...
+        $conference = Conference::findOrFail($request->input('conferenceId'));
+
+        $submission = Submission::create([
+            'conference_id' => $conference->id,
+            'talk_revision_id' => $talk->id
```
mattstauffer: Trailing comma please

app/Http/Controllers/SubmissionsController.php
```diff
@@ -2,32 +2,31 @@
...
+        ]);
+        return response()->json([
+            'status' => 'success',
+            'message' => 'Talk Submitted',
+            'submissionId' => $submission->id
```
mattstauffer: Trailing comma please

app/Talk.php
mattstauffer: 👌

app/UuidBase.php
mattstauffer: This is still open -- please either move this to modelEvents, or delete that file, so we're not double-binding this on those.

database/migrations/2019_02_04_214119_create_acceptances_table.php
mattstauffer: 👌

database/migrations/2019_02_05_083929_add_primary_key_to_submissions.php
mattstauffer: Please import App\Submission and add space after function and before opening brace

database/migrations/2019_02_05_083929_add_primary_key_to_submissions.php
mattstauffer: Please import Uuid

database/migrations/2019_02_05_084300_add_acceptance_id_to_submissions.php
mattstauffer: Please drop leading space

database/migrations/2019_02_05_084300_add_acceptance_id_to_submissions.php
mattstauffer: Please drop cascade

database/migrations/2019_02_05_084300_add_acceptance_id_to_submissions.php
mattstauffer: Please drop trailing llines

database/migrations/2019_02_05_113705_drop_status_column_from_submissions.php
mattstauffer: Please drop leading space

app/Http/Controllers/ConferencesController.php
greydnls: I can update the template. 👍

greydnls:
@mattstauffer I think you might be looking at an outdated diff? Most of your most recent comments are showing as outdated and have been addressed.

app/Http/Controllers/AcceptancesController.php
greydnls:
```
 $userHasAccess = auth()->user()->talks->filter(function ($talk) use ($submission) {
            return $talk->talkRevision === $submission->talkRevision;
        }) !== false;
        if (!$userHasAccess) {
            throw new Exception("Not Authorized");
        }
```

That's what I did, but I think your suggestion is probably a better one.

mattstauffer:
image
BOOO GITHUB BOOO... i refreshed several times because you had marked them as fixed but they were still showing. OK, take two.

app/Http/Controllers/AcceptancesController.php
mattstauffer: Yah.. I didn't see yours because of the outdated diff. Sigh. Let me take a look.

mattstauffer: Yah I think mine's a bit clearer but also makes less database calls so let's go with that --thanks!

app/Transformers/TalkForConferenceTransformer.php
```diff
@@ -0,0 +1,28 @@
...
+        $currentTalk = $talk->current();
+
+        $submission = $talk->getMySubmissionForConference($conference);
+        $acceptance = $submission ? $submission->acceptance : null;
+
+        return [
+            'id' => $talk->id,
+            'title' => $currentTalk->title,
+            'url' => $currentTalk->getUrl(),
+            'submitted' => $acceptance ? false : !!$submission,
```
mattstauffer: Same note I gave here in the bad diff--I think I'd like to think of an accepted talk as still submitted as well.

database/factories/Factory.php
```diff
@@ -57,3 +57,25 @@
...
     ];
 });
+
+$factory->define(App\Submission::class, function () {
+    return [
+        'talk_revision_id' => function () {
+            return factory(App\TalkRevision::class)->create()->id;
+        },
+        'conference_id' => function () {
+            return factory(App\TalkRevision::class)->create()->id;
```
mattstauffer: This should be Conference::class

database/factories/Factory.php
```diff
@@ -57,3 +57,25 @@
...
+    ];
+});
+
+$factory->define(App\Acceptance::class, function () {
+    return [
+        'talk_revision_id' => function () {
+            return factory(App\TalkRevision::class)->create()->id;
+        },
+        'conference_id' => function () {
+            return factory(App\TalkRevision::class)->create()->id;
```
mattstauffer: This should be Conference::class

database/migrations/2019_02_05_084300_add_acceptance_id_to_submissions.php
```diff
@@ -0,0 +1,36 @@
...
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('submissions', function (Blueprint $table) {
+            $table->string('acceptance_id', 36)->nullable();
```
mattstauffer: Interesting.. I sort of think of the acceptance as belonging to the submission instead of the other way around. What do you think?

resources/assets/js/components/TalksOnConferencePage.vue
```diff
@@ -50,55 +72,76 @@ export default {
...
-                // as axios intends it
-                axios.delete('/submissions', {params: data})
-                    .then(response => {
-                        talk.loading = false;
-                    })
-                    .catch(e => {
-                        alert('Something went wrong.');
-                        talk.loading = false;
-                    });
+            switch(status){
```
mattstauffer: spaces after switch and before brace, please

database/migrations/2019_02_05_084300_add_acceptance_id_to_submissions.php
greydnls: I was thinking that most often, you'd want to know if a submission was accepted, rather than what submission an acceptance came from. 

It allows for easy scopes on submissions for `where acceptance_id is not null`, without having to join to acceptances table or do multiple queries. It's easy enough to change if you feel strongly in the other direction, though.

mattstauffer: Makes sense! Let's try it :)

greydnls:
Is `npm run prod` something I need to do as part of this PR or is it something that is done as part of deployment?

mattstauffer:
Deployment. Thanks for asking. :)

mattstauffer:
Ready for one more review?

greydnls:
Yes please!

app/Http/Controllers/AcceptancesController.php
```diff
@@ -0,0 +1,39 @@
...
+        if (auth()->user()->id != $submission->talkRevision->talk->author_id) {
+            return response('', 401);
```
mattstauffer: Is this better than `abort(401)` in any way? I honestly don't know :)

app/Http/Controllers/AcceptancesController.php
```diff
@@ -0,0 +1,39 @@
...
+        $acceptance = Acceptance::createFromSubmission($submission);
+
+        return response()->json([
+            'status' => 'success',
+            'message' => 'Talk Accepted!',
+            'acceptanceId' => $acceptance->id,
+        ]);
+    }
+
+    public function destroy($id)
```
mattstauffer: Could we use route model binding here and save a line?

```php
    public function destroy(Acceptance $acceptance)
    {
```

app/Http/Controllers/AcceptancesController.php
```diff
@@ -0,0 +1,39 @@
...
+    {
+        $acceptance = Acceptance::findOrFail($id);
+        if (auth()->user()->id != $acceptance->submission->talkRevision->talk->author_id) {
```
mattstauffer: Not in this PR--I'd love to just get this merged--but I think we can move these into native Laravel ACL and drop some code here.

app/Http/Controllers/AcceptancesController.php
```diff
@@ -0,0 +1,39 @@
...
+            return response('', 401);
+        }
+
+        $acceptance->submission->removeAcceptance();
```
mattstauffer: What if this were a hook on `submission->delete()`? Any time I delete a submission, null out the acceptance_id on its related submission?

app/Http/Controllers/AcceptancesController.php
```diff
@@ -0,0 +1,39 @@
...
+        $acceptance->delete();
+
+        return response()->json(['status' => 'success', 'message' => 'Talk Moved back to Submitted Status']);
```
mattstauffer: Can we change this message to "Talk un-marked as accepted."  .. I still want to think of "accepted" and "submitted" as not mutually exclusive.

app/Http/Controllers/SubmissionsController.php
```diff
@@ -2,32 +2,38 @@
...
-    public function destroy(Request $request)
+    public function destroy($id)
     {
-        $talk = auth()->user()->talks()->findOrFail($request->input('talkId'));
-
-        $this->dispatch(new DestroySubmission(
-            $request->input('conferenceId'),
-            $talk->id
-        ));
+        $submission = Submission::findOrFail($id);
```
mattstauffer: Route model binding here too

database/factories/Factory.php
```diff
@@ -45,6 +45,9 @@
         'slides' => 'http://speakerdeck.com/mattstauffer/the-best-talk-ever',
         'description' => 'The best talk ever!',
         'organizer_notes' => 'No really.',
+        'talk_id' => function () {
```
mattstauffer: 👌

database/migrations/2019_02_05_083929_add_primary_key_to_submissions.php
```diff
@@ -0,0 +1,41 @@
...
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('submissions', function (Blueprint $table) {
+            $table->uuid('id')->default(0);;
```
mattstauffer: Double semicolon here

database/migrations/2019_02_05_083929_add_primary_key_to_submissions.php
```diff
@@ -0,0 +1,41 @@
...
+            $table->primary('id');
+        });
+
+        Submission::all()->map(function ($submission){
+            $submission->id = Uuid::uuid4();
+            $submission->save();
+        });
```
mattstauffer: Should we modify it to not have a default after this?

app/Http/Controllers/AcceptancesController.php
greydnls: When I was working on it `abort(401)` didn't return a 401 API response like I expected, it was throwing a 500 error, changing to this stopped that.

app/Http/Controllers/AcceptancesController.php
greydnls: I agree, I don't think this is very elegant. I need to read more about Laravel ACL, but I can do a fast follow PR to address and clean that up. :+1:

app/Http/Controllers/AcceptancesController.php
greydnls: I'm a little confused. I think what you're mean is add a hook so that any time an acceptance is deleted it removes the `acceptance_id` on it's corresponding submission?

database/migrations/2019_02_05_083929_add_primary_key_to_submissions.php
greydnls: :+1:

greydnls:
@mattstauffer Okay, I think this is ready for another round. Thank you for your patience!

app/Http/Controllers/AcceptancesController.php
mattstauffer: Whoops! Yes. I meant, could you hook on delete acceptance, to null out the acceptance_id on its connected acceptance.

app/Acceptance.php
```diff
@@ -0,0 +1,44 @@
...
+
+    protected $fillable = [
+        'talk_revision_id',
+        'conference_id',
+    ];
+
+    protected static function boot()
+    {
+        parent::boot();
+        static::deleting(function (Acceptance $acceptance) {
```
mattstauffer: 👌

app/Http/Controllers/AcceptancesController.php
mattstauffer: Weird. That shouldn't happen but we can test later.

mattstauffer:
Putting it up on staging now!

PR #245, opened 2019-02-08
Stop compiled js/css from being committed during development

greydnls:
When I run `npm run dev` while developing it changes these files. Since those files are compiled during deployment I thought it would be a good idea to stop changes from happening in these directories.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,5 @@
 .env
 .sass-cache/
 /storage/*.key
+/public/css
+/public/js


mattstauffer:
I was wrong. Turns out it doesn't. Deploy script:

```
cd /home/forge/symposiumapp.com
git pull origin master
composer install --no-interaction --prefer-dist --optimize-autoloader
echo "" | sudo -S service php7.1-fpm reload

if [ -f artisan ]
then
    php artisan migrate --force
fi
```

I'd love to add it eventually so we could actually be building the prod assets, but I'm a bit wary of doing an `npm run prod` and taking the site down for how long that takes on every deploy. Maybe it's worth moving the site over to Envoyer so we can get a full Capistrano-style deploy? hmmm

mattstauffer:
@greydnls i changed my mind. it's like this on develop now. :) thanks!

PR #246, opened 2019-02-14
Am feature/summary cfp email and default to page link

andrewmile:
This PR does two things:

1. Tests are now run without connecting to Algolia, which gives a bit of a speed increase.
2. When the `symposium:notifyCfps` command is run, if there are multiple conferences with open CFPs, they will be listed in one email instead of sending an email for each one.

### DIFF
--- a/app/Console/Commands/SendNotificationForOpenCFPs.php
+++ b/app/Console/Commands/SendNotificationForOpenCFPs.php
@@ -31,16 +31,19 @@ class SendNotificationForOpenCFPs extends Command
      */
     public function handle()
     {
-        $conferences = Conference::approved()->notShared()->openCFP()->get();
+        $conferences = $this->conferencesToShare()->get();
 
         if ($conferences->isEmpty()) {
             return;
         }
 
-        $conferences->each(function ($conference) {
-            $conference->update(['is_shared' => true]);
-        });
+        $this->conferencesToShare()->update(['is_shared' => true]);
 
         Notification::send(User::wantsNotifications()->get(), new CFPsAreOpen($conferences));
+    }
+
+    protected function conferencesToShare()
+    {
+        return Conference::approved()->notShared()->openCFP();
     }
 }
--- a/app/Console/Commands/SendNotificationForOpenCFPs.php
+++ b/app/Console/Commands/SendNotificationForOpenCFPs.php
@@ -2,10 +2,10 @@
 
 namespace App\Console\Commands;
 
-use App\User;
 use App\Conference;
-use Illuminate\Console\Command;
 use App\Notifications\CFPsAreOpen;
+use App\User;
+use Illuminate\Console\Command;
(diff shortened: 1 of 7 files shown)


app/Console/Commands/SendNotificationForOpenCFPs.php
```diff
@@ -31,10 +31,16 @@ class SendNotificationForOpenCFPs extends Command
...
-            $conference->update(['is_shared' => true]);
+        $conferences = Conference::approved()->notShared()->openCFP()->get();
+
+        if ($conferences->isEmpty()) {
+            return;
+        }
 
-            Notification::send(User::wantsNotifications()->get(), new CFPIsOpen($conference));
+        $conferences->each(function ($conference) {
+            $conference->update(['is_shared' => true]);
```
mattstauffer: This is a minor nitpick, but could we make this a single update?

```php
Conference::approved()->notShared()->openCFP()->update(['is_shared' => true]);
```

I know it's duplicating the query, but this way it's a single SQL call, which I like. What do you think?

app/Notifications/CFPsAreOpen.php
```diff
@@ -2,28 +2,28 @@
...
-    public $conference;
+    public $conferences;
 
     /**
      * Create a new notification instance.
      *
      * @param Conference $conference
      */
-    public function __construct(Conference $conference)
+    public function __construct(Collection $conferences)
```
mattstauffer: What if we drop this type hint so you could also pass an array in? Not that we need to at the moment, but I don't see a need to type hint it and lock it down, since the Blade template is just foreaching anyway.

app/Notifications/CFPsAreOpen.php
```diff
@@ -2,28 +2,28 @@
...
     /**
      * @var Conference
      */
-    public $conference;
+    public $conferences;
 
     /**
      * Create a new notification instance.
      *
      * @param Conference $conference
```
mattstauffer: Either fix or delete this docblock please

app/Console/Commands/SendNotificationForOpenCFPs.php
andrewmile: That's a good idea.  I updated it so it is now a single update query.  Rather than duplicating the whole query, I hid the common part of the query behind a `conferencesToShare` private method.

app/Notifications/CFPsAreOpen.php
andrewmile: I just removed it since it was really telling nothing more than the property name told anyway.

app/Notifications/CFPsAreOpen.php
andrewmile: That makes sense.  I went ahead and deleted the type hint.

andrewmile:
@mattstauffer This is ready for your review.

mattstauffer:
👍

PR #248, opened 2019-02-25
Allow users to dismiss conferences

andrewmile:
This PR builds on PR #215, which replaced the favorites feature with a dismissing feature.  In this PR, I have brought back reenabled favorites so we will have the ability to favorite or dismiss a conference.

## Favoriting and Dismissing Conferences

Conferences that have not yet been favorited or dismissed will display both the star (to favorite) and X (to dismiss) action buttons, as seen below:

both-actions-view

## Dismissed Conferences

Dismissed conferences will not appear in any filter except the "dismissed" filter.  Once a conference is dismissed, it cannot be favorited until it has been restored.  Dismissed conferences can be restored by clicking the plus action button by the conference listing within the "dismissed" filter as seen below:

dismissal

## Favorite Conferences

Favoriting conferences continues to work as before, by toggling the star action button to favorite or unfavorite a conference as seen below.  Conferences that have been favorited cannot be dismissed until the conference has been un-favorited.

favorite

## Viewing Conferences as a Guest

In addition, action buttons no longer appear beside conference listings when viewing conferences as a guest, as seen below:

guest-view


app/User.php
```diff
@@ -123,6 +128,7 @@ protected static function boot()
             }
 
             \DB::table('favorites')->where('user_id', $user->id)->delete();
+            \DB::table('dismissed_conferences')->where('user_id', $user->id)->delete();
```
damiani: Import `DB` rather than referencing `\DB`

resources/assets/sass/app.scss
```diff
@@ -267,25 +267,27 @@ legend {
...
         margin-top: 0.25em;
     }
 
-    .fav-button {
+    .conference-actions {
+        text-align: center;
+        margin-right: 1rem;
+        display: flex;
+        flex-direction: column;
+    }
```
damiani: Alphabetize CSS properties

resources/views/conferences/index.blade.php
```diff
@@ -41,43 +41,50 @@ class="glyphicon glyphicon-plus"
...
-                            @endif
+                            <div class="conference-actions">
+                                @if (Auth::check() && !$conference->isFavorited())
+                                    @if ($conference->isDismissed())
+                                        <a href="/conferences/{{ $conference->id  }}/undismiss" class="action-button" title="I am interested in this conference"><span class="glyphicon glyphicon-plus"></a>
+                                    @else
+                                        <a href="/conferences/{{ $conference->id  }}/dismiss" class="action-button" title="I am not interested in this conference"><span class="glyphicon glyphicon-remove"></a>
+                                    @endif
+                                @endif
+                                @if (Auth::check() && !$conference->isDismissed())
```
damiani: Our general rule in HTML is to include an extra linebreak between adjacent elements, including Blade control structures. So this would be:

```
...
@endif 

@if
... 
```

and elsewhere, there should be linebreaks between things like:

```
...
</div>

<div class="details">
...
```

resources/views/conferences/index.blade.php
```diff
@@ -41,43 +41,50 @@ class="glyphicon glyphicon-plus"
...
+                                        </span>
+                                    </i></p>
+                                @if (Auth::check() && $conference->appliedTo())
+                                    <b>Already Sent Proposal</b>
+                                @endif
 
-                            <?php /* TODO: cleaner substr */ ?>
-                            <p>{{ mb_substr($conference->description, 0, 100) }}
-                                ...</p>
+                                <?php /* TODO: cleaner substr */ ?>
```
damiani: Replace this with a Blade comment

```
{{-- TODO: cleaner substr --}}
```

tests/AccountTest.php
```diff
@@ -202,13 +206,94 @@ function deleting_a_user_deletes_its_associated_entities()
...
         $this->dontSeeInDatabase('favorites', [
             'user_id' => $user->id,
             'conference_id' => $favoriteConference->id,
         ]);
     }
+
+    /** @test */
+    function users_can_dismiss_a_conference()
+    {
+        $user = factory(App\User::class)->create();
```
damiani: No need for `App\` (here and in several instances below)

app/Conference.php
```diff
@@ -173,11 +181,16 @@ public function getLinkAttribute()
     }
 
     /**
-     * Get all users who favorited this conference
+     * Get all users who dismissed this conference
```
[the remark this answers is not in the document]
andrewmile: Thanks Keith, I have removed it.

app/Conference.php
```diff
@@ -127,6 +127,14 @@ public function scopeApproved($query)
         return $query->where('is_approved', true);
     }
 
+    public function scopeUndismissed($query)
+    {
+        return $query
+            ->whereDoesntHave('usersDismissed', function ($query) {
+                $query->where('id', Auth::id());
```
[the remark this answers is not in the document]
andrewmile: Yeah, I prefer the helper as well.  I just replaced all of the `Auth` facade calls in the `Conference` model with calls to the helper.

app/Conference.php
```diff
@@ -173,11 +181,16 @@ public function getLinkAttribute()
...
-    public function usersFavorited()
+    public function usersDismissed()
+    {
+        return $this->belongstoMany(User::class, 'dismissed_conferences')->withTimestamps();
+    }
+
+    public function isDismissed()
     {
-        return $this->belongstoMany(User::class, 'favorites')->withTimestamps();
+        return Auth::user()->dismissedConferences->contains($this->id);
```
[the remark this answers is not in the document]
andrewmile: Replaced with the helper.

app/Conference.php
```diff
@@ -173,11 +181,16 @@ public function getLinkAttribute()
     }
 
     /**
-     * Get all users who favorited this conference
+     * Get all users who dismissed this conference
      */
-    public function usersFavorited()
+    public function usersDismissed()
```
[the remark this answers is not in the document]
andrewmile: Moved to bottom of model relationships.

app/Http/Controllers/ConferencesController.php
```diff
@@ -185,8 +188,34 @@ public function destroy($id)
         return redirect('conferences');
     }
 
+    public function dismiss($conferenceId)
+    {
+        $conference = Conference::findOrFail($conferenceId);
```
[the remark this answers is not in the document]
andrewmile: Replaced with inline query.

app/Http/Controllers/ConferencesController.php
```diff
@@ -185,8 +188,34 @@ public function destroy($id)
...
+        auth()->user()->dismissedConferences()->detach($conferenceId);
+
+        return redirect()->back();
+    }
+
     public function favorite($conferenceId)
     {
+        $conference = Conference::findOrFail($conferenceId);
+
+        if ($conference->isDismissed()) {
```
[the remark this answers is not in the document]
andrewmile: Replaced with inline query.

app/User.php
andrewmile: I imported the facade.

resources/assets/sass/app.scss
andrewmile: This is fixed.

resources/views/conferences/index.blade.php
andrewmile: Got it, line breaks added.

resources/views/conferences/index.blade.php
andrewmile: Done

tests/AccountTest.php
andrewmile: Gotcha, yeah I wasn't sure what our convention was since there was some of each already here.  I prefer to use the class import as well.

app/Http/Controllers/ConferencesController.php
```diff
@@ -185,8 +188,30 @@ public function destroy($id)
         return redirect('conferences');
     }
 
+    public function dismiss($conferenceId)
```
josecanhelp: These controller methods would be a good place to use [implicit route model binding](https://mattstauffer.com/blog/implicit-model-binding-in-laravel-5-2/).

It would look something like this:
image

You would also need to change the routes in `web.php` to use `{conference}` instead of `{id}`.

resources/views/conferences/index.blade.php
```diff
@@ -41,44 +41,58 @@ class="glyphicon glyphicon-plus"
...
+                                        <br>CFP: <b>{{ $conference->cfpStartsAtDisplay() }}</b> to <b>{{ $conference->cfpEndsAtDisplay() }}</b>
+                                        </span>
+                                    </i>
+                                </p>
+
+                                @if (Auth::check() && $conference->appliedTo())
+                                    <b>Already Sent Proposal</b>
+                                @endif
+
+                                {{-- TODO: cleaner substr --}}
```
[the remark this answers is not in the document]
josecanhelp: Nevermind, I see it's been there for a long time.

app/Http/Controllers/ConferencesController.php
josecanhelp: This doesn't need to happen for this PR. We could add a new task later to use implicit route model binding.

PR #251, opened 2019-03-16
Update Installation section in Readme

cookieguru:
1. Move the Algolia key section immediately after the `.env` step.  A key is necessary to run `php artisan migrate --env=local --seed` for the first time 🤷🏻‍♂️.  Not sure if a key is necessary to run the migrations without seed data, but in any case it's helpful to have that step first (though a hard dependency on Algolia isn't ideal).
2. Add steps to install frontend dependencies and generate CSS.  The site "works" but looks unpretty without this step.


mattstauffer:
Thanks again @cookieguru !

PR #258, opened 2019-03-25
Add a Try Catch to the JoindIn Client Call

josecanhelp:
Currently, if the Joind.in client call has a server error, we don't catch the exception. This adds the try/catch block.

### DIFF
--- a/app/Console/Commands/SyncJoindInEvents.php
+++ b/app/Console/Commands/SyncJoindInEvents.php
@@ -2,17 +2,17 @@
 
 namespace App\Console\Commands;
 
+use App\JoindIn\ConferenceImporter;
+use Exception;
 use Illuminate\Console\Command;
 use JoindIn\Client as JoindInClient;
-use App\JoindIn\ConferenceImporter;
 
 class SyncJoindInEvents extends Command
 {
     protected $name = 'joindin:sync';
-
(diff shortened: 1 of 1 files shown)


app/Console/Commands/SyncJoindInEvents.php
```diff
@@ -30,11 +30,17 @@ public function handle()
...
+            $events = $this->client->getEvents();
 
-        $this->info('Events synced.');
+            foreach ($events as $event) {
+                $this->info('Creating/updating event ' . $event['name']);
+                $this->importer->import($event['id']);
+            }
+
+            $this->info('Events synced.');
+        } catch (Exception $exception) {
```
mattstauffer: Are we intentionally catching exceptions on lines 36-39? If not, would it be better in any way to just try around line 34? I'm never 100% sure how I should handle this.

josecanhelp: I had the same thought. I think it depends. In this case, the catch would also grab any thrown exceptions from calling the importer. I don't necessarily think that's what we want. Since we only want to silence the github client calls. I'll fix this up.

josecanhelp: I meant the JoindIn client calls.

PR #260, opened 2019-04-07
Fix command for running tests

cookieguru:
Output for `composer test`:
```plain
> ./vendor/bin/phpunit
'.' is not recognized as an internal or external command,
operable program or batch file.
Script ./vendor/bin/phpunit handling the test event returned with error code 1
```

The `./` part is not necessary on any platform, and the `/` cause problems on Windows, and in any case, none are required due to Composer's [PATH handling](https://getcomposer.org/doc/articles/scripts.md#writing-custom-commands)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -66,7 +66,7 @@
             "Illuminate\\Foundation\\ComposerScripts::postAutoloadDump",
             "@php artisan package:discover"
         ],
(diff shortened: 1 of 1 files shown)


mattstauffer:
Great catch. Thanks @cookieguru!

PR #262, opened 2019-04-08
Add .idea folder to gitignore

cookieguru:


mattstauffer:
Thanks! Our policy is to put machine-specific config in global git ignores and to reserve projects' git ignores for project-specific files.

Here's how, if you or anyone else are curious:

https://twitter.com/stauffermatt/status/939256695783612416

cookieguru:
Thing is, some projects _do_ check that in when all developers are using the same IDE so I don't want that globally on my system.

mattstauffer:
👌 Then if you're working on a project like Symposium, you may end up having to do something like this:

https://twitter.com/stauffermatt/status/939149586467352576

cookieguru:
What's your aversion to having them in to project's ignore file?

mattstauffer:
Because they're not project-related. PHPstorm has nothing to do with Symposium. Neither does Sublime Text, MacVim, Coda, VS Code, or any other editor an individual developer chooses to use. Neither do `.DS_Store` files, or any files auto-generated by Windows.

In my opinion, the project-specific `.gitignore` should ignore files for that project and nothing else.

PR #265, opened 2019-04-19
Update dependencies to avoid security issues

josecanhelp:


package.json
```diff
@@ -10,12 +10,17 @@
...
-    "vue": "^2.1.10"
+    "cross-env": "^5.1",
+    "jquery": "^3.2",
+    "laravel-mix": "^4.0.7",
+    "lodash": "^4.17.5",
+    "popper.js": "^1.12",
+    "resolve-url-loader": "^2.3.1",
+    "sass": "^1.15.2",
+    "sass-loader": "^7.1.0",
+    "vue": "^2.5.17",
```
damiani: Build fails because `vue` and `vue-template-compiler` don't match; this can be bumped to `^2.6.10` and it will build successfully.

package.json
```diff
@@ -10,12 +10,17 @@
...
     "bootstrap-sass": "^3.3.7",
-    "cross-env": "^5.0.1",
-    "jquery": "^3.1.1",
-    "laravel-mix": "^1.0",
-    "lodash": "^4.17.4",
-    "vue": "^2.1.10"
+    "cross-env": "^5.1",
+    "jquery": "^3.2",
+    "laravel-mix": "^4.0.7",
+    "lodash": "^4.17.5",
```
damiani: This isn't used anywhere as far as I can tell.

package.json
```diff
@@ -10,12 +10,17 @@
...
+    "popper.js": "^1.12",
```
damiani: This isn't used anywhere as far as I can tell.

package.json
```diff
@@ -10,12 +10,17 @@
...
-    "axios": "^0.16.2",
+    "axios": "^0.18",
     "bootstrap-sass": "^3.3.7",
-    "cross-env": "^5.0.1",
-    "jquery": "^3.1.1",
-    "laravel-mix": "^1.0",
-    "lodash": "^4.17.4",
-    "vue": "^2.1.10"
+    "cross-env": "^5.1",
+    "jquery": "^3.2",
```
damiani: There's a moderate vulnerability that goes away w/ `^3.4`, is there any issue using that version of jQuery?

Issue #161, opened 2016-03-21
Use Open Community Data


EmiiKhaos:
I don't know how conferences are populated into Symposium, but it would be neat to use Open Community Data: https://github.com/afilina/dev-community-data

mattstauffer:
Right now it's manual and then also pulls in Joind.In. It would be very nice to pull in as many sources of data as possible.

michaelrtm:
Hey @mattstauffer, would the starting point for this be looking at creating something like an `ImportConference` contract that had maybe `import`, `create` and `update` methods? I think I'd then rewrite the `JoindIn/ConferenceImporter` to implement that contract, instead of having method names like `updateConferenceFromJoindInEvent`, then start on actually implementing this?

Checking if this would be the correct starting point, as I've never implemented something like this before.

mattstauffer:
@michaelrtm Sure, although I don't know if I trust a contract, because the parameters of the sort of data we'd be getting in would be different for each.

Each importer is responsible for taking data from an external service and transforming it and inserting or updating it into our database. 

It may make sense to have a single method like `sync` that each offers, and if we need something else (e.g. forcing an individual conference ID to update from its source) we could do that--but only if there's a practical need for it.

Beware of abstractions and interfaces for the sake of abstraction and interface; rather, create them when you see the need for them.

PR #268, opened 2019-05-16
Remove JoindIn client in favor of CallingAllPapers

cookieguru:
This is my attempt at fixing #161.  Using Open Community Data would be great, however that repo is pretty stagnant.  There have only been 10 additions in the last 12 months and there have certainly been more conferences than that in the same timeframe.

Enter [Calling All Papers](https://callingallpapers.com/) which is maintained by the fine folks from @JoindIn.  It does what was requested: Pull in data from as many sources as is possible, including the current source of JoindIn.  The only conference listing site that I've hard of that Calling All Papers is missing is Sessionize, but they don't have an API [and won't add one](https://twitter.com/SessionizeCom/status/1117810496382566400).

I think the biggest downside is that this (retains the current functionality) of making a second importer hard.  What if two APIs had the same conference?  How do you prevent both importers from adding the same event to the database?  I don't think this really needs to be addressed here, but would need to be in the future.  De-duping by URL is not viable since some conferences use the same URL for multiple years.  But I digress.

And finally, this makes the site feel less like a JoindIn wrapper.  I removed the jump off points to JoindIn since they'd be difficult to retain.  I'm not sure how those were used (to me they had no value), and in any case, it brings the conferences imported from JoindIn in parity with the manually-added ones.


mattstauffer:
@cookieguru Thank you for your hard work on this! I look forward to looking through this.

mattstauffer:
@cookieguru Would you mind adding me as a collaborator to your fork?  Hoping to make some modifications on your PR, and this is the best I can find about how best to do that on a fork:

https://help.github.com/en/articles/committing-changes-to-a-pull-request-branch-created-from-a-fork

cookieguru:
Done!

svpernova09:
Speaking as a member of the Joindin Leadership team: CallingAllPapers is essentially operated by @heiglandreas The rest of the Joindin team has little to do with it.

My personal preference and bias would be to see Symposium continue to pull from Joind.in as a data source, while adding any other resources that would add content and work through whatever "duplicate event scenario" that may arise.

cookieguru:
@svpernova09 a "duplicate event scenario" is guaranteed.  I'm not sure why you are advocating for duplication of work; is retirement of Calling All Papers in the Joindin road map?

heiglandreas:
@svpernova09 Why do the same work twice? CallingAllPapers is a part of the joind.in project. At least it was up to now. The fact that I'm the only one working on CallingAllPapers is easy to solve. I'm happy to see others work on it as well or even take over the project.

heiglandreas:
@cookieguru  I am working on a way to import the publicly available CfPs. There is a hidden gem that allows that ;-)

svpernova09:
> Why do the same work twice? CallingAllPapers is a part of the joind.in project. At least it was up to now. The fact that I'm the only one working on CallingAllPapers is easy to solve. I'm happy to see others work on it as well or even take over the project.

I wasn't advocating for anything other than keeping Joind.in as a data source. I have no problems at all with CallingAllPapers being a part of joindin, I was just pointing out that AFAIK you (Andreas) was the only one actively working on it.

cookieguru:
@svpernova09 Calling All Papers is stable, so I don't think much work needs to be done on it.  And I think @heiglandreas is aware that I will help out in any way (I have two PRs open right now)

mattstauffer:
Thanks, everyone. Sounds to me like Calling All Papers is our best option moving forward. I got stuck reviewing this PR and have to add some other Passport-related tooling to Symposium to make this an easier PR to review, so it's gonna be a bit, but I love the idea of relying on Calling All Papers as an aggregator to lessen our responsibility.

mcneely:
@svpernova09 Can correct me if I say anything incorrect but CAP is @heiglandreas project, if he were to be hit by a bus tomorrow, there's little interest from the rest of the team to support it at this juncture. He can do whatever he wants with it as there isn't any official over site regarding the project. Moving your source of truth to an intermediary is never advisable and could result in headaches down the road.

cookieguru:
@mcneely and @svpernova09 What is it about the Calling All Papers project that makes you not want to touch it with a 10 foot/3 meter pole?

Also I don't understand your aversion to Symposium using an open-source aggregation API.

mcneely:
@cookieguru The new team's focus is to shoring up and improving upon the main product. It's a matter of resource allocation and priority.

mattstauffer:
Sounds like CallingAllPapers should split out into its own project and develop its entirely separate set of developers so that it can be supported if @heiglandreas is ever unable or uninterested to keep it going.

heiglandreas:
See https://github.com/joindin/callingallpapers.com/issues/55

mattstauffer:
@cookieguru OK, made a few changes to master.. back at this. This looks great. My only remaining concern is: how do we handle migrating existing conferences over to their correct CAP ID?

Once we solve that I'm ready to do this.

heiglandreas:
The CAP-ID of a CfP is the sha1-hash of the final event-URI (the event-uri CAP returns). The Event URI is retrieved by following all redirects for the Event-URI and using the final destination. For details see https://github.com/joindin/callingallpapers-api/blob/master/src/Service/CfpFactory.php#L131-L155

Perhaps that helps solve that issue?

mattstauffer:
@heiglandreas I think the biggest concern is being sure that the event-URI we use to look it up (from our old JoindIn record) is the same one that CAP was using.

Does a map between CAP and JoindIn IDs or URLs or anythihng exist anywhere?

Thanks! this is super helpful.

heiglandreas:
Nope. There is no mapping between the joind.in-URL and CAP-url available. Other than that CAP takes the event-URI from joind.in and retrieves the hash in the described way from it.

mattstauffer:
@heiglandreas OH, but the event-URI that CAP calculates its ID based off of is consistently pull-able from JoindIn records? If so, we can build a one-time translator. 👌

heiglandreas:
Weeelllllll. As long as the people doing the CfP do not change their "final" event-URI during the time the CfP is open that should be consistent. :grin:

Oh. Keep in mind, that the final '/' of the Event URI is removed before calculating the SHA1.

cookieguru:
@heiglandreas How does that work with conferences that use the same URL from year to year?  E.g. this year's Dutch PHP conference is `https://www.phpconference.nl/` (with the last `/` removed) as was last year's.  Some conferences put the year in the URL, so those should de-dupe easily.

@mattstauffer If you want to send me a DB dump of prod I can write a migration script.

heiglandreas:
In that case there is not a new entry created for the CfP but the old one is updated with the new information. As long as there are no events that overlap and share the same URL it will be unique for the duration of the CfP. And that was the only relevant thing for CAP. In Symposium that might indeed mean that a historic conference might link to a current CfP...

cookieguru:
@mattstauffer In light of this, I changed the logic to [append the year to the ID](https://github.com/tightenco/symposium/pull/268/commits/24deacd88301d626826f91fbf6ed39ea2a6c842a#diff-d63183a5a193bb494fcc4b60ecd71d20R50) that Symposium stores.  The ID is defined as the year the conference starts.  If that year is 1970, it falls back to the year the CFP ends.  In the unlikely case that is also 1970, it throws.

I also lost your code style fix commit when I fast-forwarded my branch to Symposium's `develop` forgetting to pull down the calling-all-papers branch from my own repo, but I worked the changes back in.

mattstauffer:
@cookieguru Good call; we'd like to eventually link the same event across multiple years some way, but for now they should be considered separate events.

I'm in the midst of a cross-country move, so please excuse my delays here. I'm going to try to give it another read through right now and see if I can get through far enough to merge this in and try it out on staging.

Thanks!

mattstauffer:
OK, this is up on https://staging.symposiumapp.com/ to play around with for a bit before I push to master. Thank you so much for your hard work @cookieguru and your input others, especially @heiglandreas!

mattstauffer:
OK, first bug: Whatever logic checks for create/update on conferences is failing. I ran the sync four times and got the same conference added four times. I'll dig into it shortly but just something we'll have to fix before this can go live.

image

mattstauffer:
It looks like the code is passing the CAP ID into firstOrNew, which seems to work fine to me in Tinker:

image

I'll see if I can do more testing locally or write some more tests to reproduce the problem later. Welcome input if you have time @cookieguru but I know this is all out of the blue time-wise, so no worries if you don't.

mattstauffer:
Ha. Turns out the ID field needed 4 more characters once you added the year:

https://github.com/tightenco/symposium/commit/cd79394c35f1356970aae09757a54f17a8522ac8

cookieguru:
🤦🏻‍♂️

PR #276, opened 2019-08-31
Fix login and signup errors

### LINKED ISSUE #275: Login and signup errors are still being misdirected
There are four places for potential errors to display: the sign in box on the home page, the sign up box on the home page, and the sign up and sign in pages. Almost all of the errors on these pages are either not showing, or not showing in the right spot. Let's fix.

https://stefanbauer.me/tips-and-tricks/multiple-forms-with-same-input-names-on-one-page

tomdiggle:
This commit fixes #275 the errors on the sign up, login and home page. It uses a seperate FormRequest for Login and Register each with a custom loginForm and registerForm ErrorBag.

### DIFF
--- a/app/Http/Controllers/Auth/LoginController.php
+++ b/app/Http/Controllers/Auth/LoginController.php
@@ -3,10 +3,8 @@
 namespace App\Http\Controllers\Auth;
 
 use App\Http\Controllers\Controller;
-use Exception;
+use App\Http\Requests\Auth\LoginFormRequest;
 use Illuminate\Foundation\Auth\AuthenticatesUsers;
-use Illuminate\Http\Request;
-use Illuminate\Support\Facades\Lang;
 
 class LoginController extends Controller
 {
@@ -23,8 +21,6 @@ class LoginController extends Controller
 
     use AuthenticatesUsers;
 
-    public $validatesRequestErrorBag = 'login';
-
     /**
      * Where to redirect users after login.
      *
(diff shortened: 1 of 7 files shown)


tests/AccountTest.php
```diff
@@ -34,10 +34,7 @@ function invalid_signups_dont_proceed()
     {
         $this->visit('register')
             ->press('Sign up')
-            ->seePageIs('register')
-            ->see('The name field is required')
-            ->see('The password field is required')
-            ->see('The email field is required');
+            ->seePageIs('register');
```
mattstauffer: Are we no longer seeing those errors?

app/Http/Controllers/Auth/LoginController.php
```diff
@@ -41,4 +37,37 @@ public function __construct()
     {
         $this->middleware('guest', ['except' => 'logout']);
     }
+
```
mattstauffer: Any time I take a method and reproduce it purely to change its syntax, I like to A) pass through as much of the functionality up to its parent as I can, and B) note that it's 98% (or whatever) the same as its parent so that, in the future, someone knows why we have all this complexity here that aren't in most apps. E.g. "This is almost the exact same method as Illuminate\Whatever\LoginControllerTraitOrWhatever@login, just modified to accept a form request" (or whatever.)

Same note for the register method on the RegisterController.

tomdiggle: I agree. I've just changed to extend the parent methods and pass a RegisterFormRequest and LoginFormRequest. Hope this solves this issue.

tests/AccountTest.php
tomdiggle: Ah, yes we do see those errors. I've re-added them and also the errors are now displayed below the text inputs.

Screenshot 2019-09-04 at 21 15 53

app/Http/Controllers/Auth/LoginController.php
mattstauffer: Brilliant--thanks!

mattstauffer:
Thanks so much @tomdiggle!

Issue #274, opened 2019-08-23
GitHub icon on login page is borked


mattstauffer:
![qwerqwerasdfa](https://user-images.githubusercontent.com/151829/63623947-038a3b80-c5c9-11e9-9f24-7183485ace5c.gif)

tomdiggle:
I'm taking a look at this issue, because only login with GitHub is supported, would it make sense to replace the current 'auth-icons.png' file with an image of only the GitHub icon? I'd also update the CSS to use tailwind.

mattstauffer:
@tomdiggle Yah, I think that makes sense :)

PR #277, opened 2019-09-08
Fix for login with GitHub button #274

tomdiggle:
This PR fixes the issue #274 with the login with Github button. The button has been redesigned using tailwind and refactored to a component so it can be easily reused. The Github icon is now an svg image inserted inline using the `nothingworks/blade-svg` package.

In the future any svg images can be inserted inline using the `nothingworks/blade-svg` package, reducing the dependency of Glyphicon Halflings and Octicons which contain a lot of icons that are not used. For example, `octicons.css` is loaded on every page but only one icon is used in `bio-in-list.blade.php`.

Tested on:
Safari 12.1.2 macOS
FireFox 68.0.2 macOS
Chrome 76.0.3809.132 macOS


mattstauffer:
Great work! Thank you!!

PR #278, opened 2019-09-20
Minor fixes

tomdiggle:
Replaced the only Octicon icon that was being used with a svg image. Therefore making Octicon package redundant. Also, a error message is now displayed when logging in the invalid credentials and the login and sign up error messages are wrapped in an `@if` statement.


resources/views/partials/log-in-form.blade.php
```diff
@@ -1,16 +1,24 @@
 {!! Form::open(['route' => 'login']) !!}
     {{ csrf_field() }}
 
+    @if ($errors->default->first('email'))
```
mattstauffer: What's the circumstance in which we have the email problem on the default error bag instead of loginForm?

tomdiggle: If the user tries to log in with the wrong email or password then for some reason, which I'm not entirely sure the error is returned using the default error bag. The request is still a `LoginFormRequest` but it doesn't use the custom `loginForm` error bag which is a bit strange.

mattstauffer: Got it--thanks!

PR #279, opened 2019-10-10
Update lodash package to resolve vulnerabilities

chrisforrence:
It appears that lodash@4.17.5 had a good number of security vulnerabilities according to npm audit. This change simply updates lodash to 4.17.15, which addresses those vulnerabilities.


mattstauffer:
Thanks Chris!

PR #281, opened 2019-11-08
Disable email registration

andrewmile:
This PR disables the ability to register for a new account with an email address by disabling the registration routes and removing the registration form and sign up link from the home page.  Visiting `/register` or making a post request to `/register` will result in a `404`.

Sign up link removed from nav bar
image

Email registration form removed from home page
image

### DIFF
--- a/routes/web.php
+++ b/routes/web.php
@@ -48,10 +48,6 @@ Route::get('log-out', ['as' => 'log-out', 'uses' => 'Auth\LoginController@logout
 // Disable email registration
 Auth::routes(['register' => false]);
 
-// Route::get('sign-up', function () {
-//     return redirect('register');
-// });
-
 Route::group(['middleware' => 'auth'], function () {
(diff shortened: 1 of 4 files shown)


routes/web.php
```diff
@@ -45,11 +45,12 @@
...
-Auth::routes();
+// Disable email registration
+Auth::routes(['register' => false]);
 
-Route::get('sign-up', function () {
-    return redirect('register');
-});
+// Route::get('sign-up', function () {
+//     return redirect('register');
+// });
```
damiani: Do we know why this route was here to begin with? I don't see it referenced anywhere in the code. If it's used by some external link (like from an email), then we should probably leave it active, and instead redirect to `'/'` or something.

andrewmile: I imagine it was from an older version of Laravel before the auth routes were included.  I went ahead and removed it since it isn't referenced anywhere in the app.

damiani (approved):

PR #282, opened 2019-11-13
Import latitude/longitude from CallingAllPapers

cookieguru:
With #280 these fields are tracked on the conference; might as well import them from CallingAllPapers even though [they're not completely reliable](https://github.com/tightenco/symposium/pull/261#issuecomment-548232112)


mattstauffer:
What if we null out any zeroes from CAP so we can have clearer data? And then maybe we can trigger a lat/long lookup if we get text from CAP but null lat/long?

PR #292, opened 2020-12-14
Add Github Actions Workflow

faxblaster:
Adds a Github Actions workflow to run PHPUnit tests. I'm including the npm build here, since we have some browser tests that were failing without it. 

This also removes related Travis configuration and workflow files, and adds a workflow status badge.


.github/workflows/test.yml
```diff
@@ -0,0 +1,36 @@
+name: Test
+
+on:
```
mattstauffer, line 3: I've started doing this config instead:

```yaml
on:
  pull_request:
  push:
    branches:
      - main
```
So we get it on all manual pushes to the primary branch, and then all PRs to any branch.

mattstauffer (changes requested):
Looks great! Only one note. Thanks!!

PR #301, opened 2021-03-17
Setting correct file permissions

peter279k:
# Changed log

- Set correct file permissions for non-executable files.


andrewmile (approved):
Thanks!

PR #305, opened 2021-07-01
Update dependabot vulerabilities.

krievley:
This pull request updates all of the outstanding vulnerability for `npm` packages except `tailwindcss` because the `@tailwindcss/custom-forms` package is currently at it's most up to date version and will not work with a higher version of `tailwindcss`.

Rissa and Kristin ran the associated tests and looked through the local site to ensure that features were still working with the updates.


.gitignore
```diff
@@ -11,3 +11,4 @@
 /public/js/location.js.map
 /storage/debugbar
 .phpunit.result.cache
+.idea/*
```
mattstauffer, line 14: @krievley We're still running into your end of line thing here; would you mind adding a newline here?

mattstauffer: Other than that, this is good to merge; please fix and merge. Thanks!

mattstauffer (changes requested):

PR #307, opened 2021-08-08
fix: public speakers list spacing

### LINKED ISSUE #303: Spacing on the public speakers list is off
image

The location should be much closer to the name of the speaker, and each speaker should be a bit further apart

scrlkx:
Fix #303

Before
localhost_8080_speakers

After
localhost_8080_speakers (1)


mattstauffer:
Thanks!

PR #308, opened 2021-09-11
fix conference speakers alignment

faissaloux:
- Fix conference speakers alignment (First block "Track talks" is a bit higher than its siblings)

# Before
Screenshot from 2021-09-11 11-40-57

# After
Screenshot from 2021-09-11 11-41-58


andrewmile (approved):
Nice catch, thank you!

PR #321, opened 2021-10-28
Create conferences with validated input

andrewmile:
This PR updates the `CreateConferenceForm` to create conferences with only validated input rather than all input.

While verifying that the edit conference flow only uses validated input, I realized that the `latitude` and `longitude` coordinates were not being saved on updates. This PR fixes this while also simplifying the update logic to use the validated input from the defined `conference_rules`.


faxblaster (approved):
Looks great! Nice improvement.

PR #327, opened 2021-11-12
Add reason to Submission response

faxblaster:
Adds the ability to provide a `reason` for a submission's acceptance or rejection.

Nov-12-2021 12-47-28

- Adds an edit icon to the `Conference` page for submitted talks
- Adds an edit page for Submissions, with an accept/reject option and text field for `reason`, and a sidebar with Conference info
- Adds migrations for the new column in the `acceptances` and `rejections` tables
- Combines logic from `AcceptanceController` and `RejectionController` to create those models and moves it into the `SubmissionController`
- Removes now unused routes, buttons and controller methods


faxblaster:
[Add ability to add a reason why a conference was accepted or rejected](https://trello.com/c/DaPckX8o/68-add-ability-to-add-a-reason-why-a-conference-was-accepted-or-rejected)

app/Http/Controllers/AcceptancesController.php
```diff
@@ -36,6 +14,6 @@ public function destroy(Acceptance $acceptance)
 
         $acceptance->delete();
 
-        return response()->json(['status' => 'success', 'message' => 'Talk un-marked as accepted.']);
+        return response()->json(['status' => 'successs', 'message' => 'Talk un-marked as accepted.']);
```
faxblaster, line 17:
```suggestion
        return response()->json(['status' => 'success', 'message' => 'Talk un-marked as accepted.']);
```

andrewmile (approved):
Great work on this!  👏   When I was testing out the flows locally, I realized the `<x-panel>` component doesn't really work well for displaying the the conference info so I removed it since submitting the form redirects back to the conference page anyway.

A few other smaller updates I added:
- Changed the `undo` buttons to `trashcan` icons
- Moved the `reason` column before the timestamp columns in the migrations
- Changed `reason` to a textarea
- Displayed the `reason` value on the `talks.show` page

PR #329, opened 2021-11-12
remove factories model property

faissaloux:
Since Laravel v8.68.0, Laravel can guess database factory model by default


andrewmile (approved):
TIL! 💪

Issue #309, opened 2021-10-01
Improve latitude/longitude imports from CAP


andrewmile:
#282 Imports the latitude and longitude coordinates from Calling All Papers but as noted [here](https://github.com/tightenco/symposium/pull/261#issuecomment-548232112), they can sometimes contain "0".

- [ ] If the latitude or longitude coordinates are equal to "0", we should set the values to null.
- [ ] If the latitude or longitude coordinates are null, we should attempt to geolocate the conference based on the `location` key.

andrewmile:
![](https://github.trello.services/images/mini-trello-icon.png) [Improve latitude/longitude imports from Calling All Papers](https://trello.com/c/UYFPustL/85-improve-latitude-longitude-imports-from-calling-all-papers)

andrewmile:
Closed by #334

PR #334, opened 2021-11-17
Add geocoding lat/long based on location

faxblaster:
Adds the `toin0u/geocoder-laravel` service provider to attempt geocoding lat/long based on provided location in cases where the provided lat/long values are invalid.

I added the library's recommended `post-update-cmd` and `post-install-cmd` commands to `composer.json`. @andrewmile @sariyana - Can ya'll think of any adverse effects this change might create?

Before we merge to production, I think we'll need to activate the Google Geocoding API for our production api key (unless that's already done).

Update: Looks like tests are failing due to the missing Google Maps key. I guess we could add one here on the repo and reference it in the actions workflow, but I'm open to other ideas.

This addresses #309


andrewmile:
[Improve latitude/longitude imports from Calling All Papers](https://trello.com/c/UYFPustL/85-improve-latitude-longitude-imports-from-calling-all-papers)

andrewmile (approved):

PR #333, opened 2021-11-16
Replace base test case extend class

andrewmile:
This PR updates the test files to make them more in line with what would be expected in a newer Laravel app by extending `Illuminate\Foundation\Testing\TestCase` instead of `Laravel\BrowserKitTesting\TestCase` in the base `TestCase`.  As a result of this PR, the tests are less dependent on the DOM and have more of a focus on HTTP requests and responses.  Test helpers and assertions have also been updated to reflect what is available on Laravel's `TestCase` and `TestResponse` classes.

The `laravel/browser-kit-testing` dependency has been removed.

The two older signup tests that were marked as skipped have also been removed since email signups were removed prior to the site redesign and have not been re-implemented since.

PR #335, opened 2021-11-17
Refactor tests

faissaloux:
I wanted to make those changes before but I noticed that you extending ```Laravel\BrowserKitTesting\TestCase``` .

But now after #333 extending ```Illuminate\Foundation\Testing\TestCase``` I can do this 👀 😄


andrewmile (approved):
These are some nice improvements, thank you! 👍

PR #337, opened 2021-11-18
refactor factories

faissaloux:


database/factories/BioFactory.php
```diff
@@ -15,4 +15,25 @@ public function definition()
             'body' => $this->faker->sentence(),
         ];
     }
+
+    public function user($user)
```
andrewmile, line 19: I think this can be removed in favor of calling `Bio::factory()->for($user)` and letting the model take care of the keys.

database/factories/ConferenceFactory.php
```diff
@@ -24,30 +24,52 @@ public function definition()
...
+    }
+
+    public function notShared()
+    {
+        return $this->state([
+            'is_shared' => false,
+        ]);
+    }
+
+    public function author($author)
```
andrewmile, line 69: Similar to the `user` method in the `BioFactory`, I'd like to use Laravel's `for` method but calling `Conference::factory()->for($user, 'author')` everywhere is kinda clunky so I think we should keep this `author` method and change the implementation to `return $this->for($user, 'author')`.  Then all that will need to change is passing in `$user` instead of `$user->id`.

database/factories/TalkFactory.php
```diff
@@ -13,4 +13,11 @@ public function definition()
             'author_id' => User::factory(),
         ];
     }
+
+    public function author($author)
```
andrewmile, line 17: Same feedback as in the `ConferenceFactory`

andrewmile (changes requested):
Hey @faissaloux these are very welcome changes!  I love the expressive methods on the factory classes 💪 .  I left a few pieces of feedback on the `user` and `author` methods but otherwise, this is good to go.  Thanks again for your contributions!

faissaloux:
@andrewmile Oow thank you for this! I didn't know this ```for``` method exist 👀

Done! 🚀

andrewmile (approved):
Look good, thank you! 🙌

Issue #310, opened 2021-10-02
Reporting site (data) issues


cookieguru:
There's a company that decided to spam Papercall with a multi-year ongoing event with no description and a CFP far in the future.  My previous attempts at contacting the owners of Papercall have gone nowhere, so I'm not bothering with this one.

Since Papercall's data is imported and automatically trusted this same company ends up spamming Symposium too.  Nowehere on the site is there anywhere to contact anyone in charge or alert of inappropriate behavior

andrewmile:
![](https://github.trello.services/images/mini-trello-icon.png) [Prevent adding conferences with a CFP end date that is after the conference start date](https://trello.com/c/Mm6huUfl/70-prevent-adding-conferences-with-a-cfp-end-date-that-is-after-the-conference-start-date)

PR #338, opened 2021-11-18
Conference Importer date validation

sariyana:
This PR adds the following validation rules for the Conference Importer:
- Validate that the CFP period is 2 years or less
- Validate that the conference duration is 2 years or less
- Validate event start is after cfp end

Addresses issue #310.


andrewmile (approved):
Great work on this! 👍

PR #342, opened 2021-11-22
Disable location lookup for local development

andrewmile:
This PR adds a `MIX_DISABLE_LOCATION_LOOKUP` environment variable that can be used to disable the Google Maps location lookup API during local development.  This prevents unnecessary API calls and an error display during development when a value is not set for the `GOOGLE_MAPS_API_KEY` environment variable.

**Before**
image

**After**
image


andrewmile:
[Disable google maps location wrapper if key is not available](https://trello.com/c/knS7JZaT/69-disable-google-maps-location-wrapper-if-key-is-not-available)

faxblaster (approved):
Nice!

PR #373, opened 2022-04-15
Add OG and Twitter meta tags

sariyana:
This PR adds the meta tags for OpenGraph and Twitter.


sariyana:
[Add OG and Twitter meta tags](https://trello.com/c/F1PvqRXB/95-add-og-and-twitter-meta-tags)

mattstauffer:
This looks good! We could also ask Nohemi specifically for an open graph image, but I'm going to merge this for now :)

PR #393, opened 2022-07-08
Globally ignore archived talks

### LINKED ISSUE #201: Needs better handling of active/archived talks

andrewmile:
This PR closes #201 by adding an `active` global scope on the `Talk` model.  As a result of this change, the `User::activeTalks` relationship has been removed since it is equivalent to `User::talks`.

The `user/{userId}/talks` API endpoint now supports including archived talks by adding `include-archived=1` in the query string.  This change includes replacing the `User::getTalksAttributes` and `User::getActiveTalksAttribute` method with a `sortByTitle` on a custom `TalksCollection`.

### DIFF
--- a/app/Collections/TalksCollection.php
+++ b/app/Collections/TalksCollection.php
@@ -1,0 +1,16 @@
+<?php
+
+namespace App\Collections;
+
+use App\Models\Talk;
+use Illuminate\Database\Eloquent\Collection;
+
+class TalksCollection extends Collection
+{
+    public function sortByTitle()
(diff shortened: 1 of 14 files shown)


sariyana (approved):
Awesome work 🥳 ⭐! LGTM!

PR #411, opened 2022-07-27
Speaker search improvements

### LINKED ISSUE #224: Make public speaker profile page more nuanced
E.g. searching for "FL" picks me up, but not "florida."
E.g. the page layout is a little janky/crammed/unclear/inconsistent:

image
image
image

It's also not clear that you can search by location, so we should make that obvious through UI or words.

It also might be nice to be able to add "within ___ miles/km of ___"

andrewmile:
This PR closes #224 by making the following improvements to the speaker search page:
- Speakers can be searched by state name in addition to abbreviation
    - This is implemented by mapping the state name to the abbreviation without needing to update the search index
- A `Reset` button has been added to reset the search results
- The search placeholder text has been updated to read `Name or Location`
- The speaker search page has been combined to a single `speakers.index` route
    - The `Search` button now refreshes the current page with the search term appended in the query string
    - The `speakers.search` route has been removed

**Querying by state name**
image

### DIFF
--- a/app/Http/Controllers/PublicProfileController.php
+++ b/app/Http/Controllers/PublicProfileController.php
@@ -15,8 +15,7 @@ class PublicProfileController extends Controller
 {
     public function index(SpeakerSearchRequest $request)
     {
-        $users = User::search($request->get('query'))
-            ->query(fn ($query) => $query->whereHasPublicProfile())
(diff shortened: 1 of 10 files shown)


sariyana (approved):
LGTM!

PR #417, opened 2022-08-12
Modernize validation

andrewmile:
This PR modernizes the validation calls in the `store` and `update` methods of the `TalksController` by replacing instantiation of the `Validator` object with a `SaveTalkRequest` form request.

In addition, the unneeded `BaseController` and and `Api\BaseController` classes have been removed.


sariyana (approved):
Looks good!

PR #421, opened 2022-08-19
Remove jQuery

andrewmile:
This PR removes jQuery and replaces dependent code with Vue implementations.

## Dismissing Notices
In app notices (success, error, etc) now use a custom `v-dismiss` Vue directive that will fade the notice out after two seconds or hide the notice when clicked.

## V-Calendar
The `pickadate` jQuery plugin has been replaced with the [v-calendar](https://vcalendar.io/datepicker.html) `datepicker` component.

image


sariyana (approved):
LGTM!

PR #446, opened 2022-11-11
Improve conference importer visibility

andrewmile:
This PR improves visibility of the conference importer with the following updates:
- Slack will be notified when the importer starts, finishes, rejects a conference or encounters an error
- Conferences that fail validation will now be persisted but marked as rejected
- Rejected conferences cannot be seen other than by admin
- A bug in `App\CallingAllPapers\Client` where the wrong `Event` class was imported has been fixed


sariyana (approved):
Looks great! Thanks for all of the hard work on this 🌟

PR #464, opened 2023-01-13
Use a separate Google Maps API key for the frontend

andrewmile:
This PR updates the frontend to use a separate Google Maps API key from the backend.


mattstauffer:
Looks good to me!

PR #473, opened 2023-02-03
Replace geocoder package with custom geocoder

andrewmile:
This PR replaces the `toin0u/geocoder-laravel` package with a custom `App\Services\Geocoder` package.

This change is in order to prevent attempts to geocode known invalid addresses by caching the address and returning from the Geocoder when an address is known to be invalid.

We opted to removing the geocoder-laravel dependency in favor of making a custom solution because we only need the latitude and longitude coordinates from the geocoding result and the custom class avoids unwanted logging of invalid addresses.


app/Services/Coordinates.php
```diff
@@ -0,0 +1,25 @@
+<?php
```
mattstauffer, line 1: I normally think of something under `App\Services` as classes that connect to external services, not POPOs. Maybe we just throw this under `App`? Or if you like it under `App\Services` because it's attached to `Geocoder`, maybe it goes under `App\Services\Geocoder`?

app/Services/Geocoder.php
```diff
@@ -2,10 +2,51 @@
...
 namespace App\Services;
 
+use App\Exceptions\InvalidAddressGeocodingException;
+use App\Services\Coordinates;
+use Illuminate\Support\Facades\Http;
+
 class Geocoder
 {
-    public function geocode($location)
+    public function geocode(string $address)
```
mattstauffer, line 11: Clean! I love it!

tests/Integration/GeocoderTest.php
```diff
@@ -0,0 +1,61 @@
...
+        $coordinates = $geocoder->geocode('1600 Pennsylvania Ave Washington, DC');
+
+        $this->assertEquals('38.8976633', $coordinates->getLatitude());
+        $this->assertEquals('-77.0365739', $coordinates->getLongitude());
+    }
+
+    /** @test */
+    function catching_invalid_addresses()
+    {
+        $this->expectNotToPerformAssertions();
```
mattstauffer, line 26: Wow, TIL

tests/Integration/GeocoderTest.php
```diff
@@ -0,0 +1,61 @@
+<?php
+
+namespace Tests\Feature;
+
+use App\Exceptions\InvalidAddressGeocodingException;
+use App\Services\Geocoder;
+use Illuminate\Support\Facades\Http;
+use Tests\TestCase;
+
+class GeocoderTest extends TestCase
```
mattstauffer, line 10: Will these tests fail if someone doesn't have a `GOOGLE_API_KEY` in their `.env`?

mattstauffer (changes requested):
Fantastic work! Just a few questions.

app/Services/Coordinates.php
andrewmile: I completely agree!  I'm not really sure why I put it in services.  Since it is acting as a value object I just moved it to `App\Casts`.

tests/Integration/GeocoderTest.php
andrewmile: Yes, it will fail without the key although it won't run as part of the whole test suite since I moved it to `tests/Integration`.  I was thinking this would be a test that rarely needs to run since its primary purpose is to ensure we are calling Google's API correctly.  Ideally we'd run this as part of the Github action but without exposing the key.  I made a Trello card to investigate options for this.

mattstauffer (approved):
Looks good!! Merge away! Thank you!

PR #488, opened 2023-05-05
Redesign conferences page

andrewmile:
This PR implements a redesign to the conferences index page.  Conference listings are now full width with filters and sorts moved to the top of the page as select inputs.  A search field has been added to filter the conference list while typing.  

When showing all conferences, results are constrained to the selected month, which can be modified by the query string or `previous` and `next` navigation arrows.  The selected month will default to the current month on page load.

When filtering by future conferences, all results will be shown.

Filters:
- All
- Open CFP
- Future CFP
- Unclosed CFP
- Favorites
- Dismissed

Sorts:
- Title
- Event Date
- CFP Opening Date
- CFP Closing Date

The page has been re-implemented using Livewire to update the state and results without page reloads.  In addition, favoriting and dismissing conferences is also now handled by Livewire.

image

**Filtering and sorting**
https://user-images.githubusercontent.com/1121383/236707551-cd419242-e026-4e1b-ac1a-daaf07532c2d.mp4

**Searching**
https://user-images.githubusercontent.com/1121383/236707756-2bf3f29b-11cd-4a27-94bc-e2b219d9d215.mp4

**Favoriting and dismissing**
https://user-images.githubusercontent.com/1121383/236707920-d9b7137e-fa67-4912-b80e-21527e14537d.mp4


app/Http/Controllers/Api/ConferencesController.php
```diff
@@ -20,12 +20,12 @@ public function index(Request $request)
                 $conferences = EloquentConference::future()->get();
                 break;
             case 'open_cfp':
-                $conferences = EloquentConference::openCfp()->get();
+                $conferences = EloquentConference::whereCfpIsOpen()->get();
```
mattstauffer, line 23: I haven't read the rest of the code, but in the original, "open" and "unclosed" aren't the same.

Open means "you can submit today", unclosed means "the end date of the CFP has not passed, but its start date ALSO may not have passed yet."

app/Http/Livewire/ConferenceList.php
```diff
@@ -0,0 +1,287 @@
...
+        'year',
+        'month',
+        'filter',
+        'sort',
+        'search',
+    ];
+
+    public function boot()
+    {
+        Builder::mixin($this->queryScopes());
```
mattstauffer, line 35: Uh,.. i dont think I've seen this pattern before. Can you share a link or something else to help me learn?!

tests/Feature/ConferenceTest.php
```diff
@@ -908,6 +988,73 @@ public function scopping_conferences_queries_where_has_cfp_start_date()
         $this->assertNotContains($conferenceB->id, $conferenceIds);
     }
 
+    /** @test */
+    public function scopping_conferences_queries_where_favorited_by_user()
```
mattstauffer, line 992: I think you intend "scoping" on these

mattstauffer (changes requested):

mattstauffer:
Great work on this Andrew!!

app/Http/Livewire/ConferenceList.php
andrewmile: I noticed it while source diving the `Macroable` trait to experiment with defining macro methods on the calling class.  The `mixin` method accepts a class, which in this case I'm using an anonymous returned by the `queryScopes` method.  The catch is that the methods in the mixin class all need to return callables so each method is one level of indentation greater than they otherwise would be.

app/Http/Controllers/Api/ConferencesController.php
andrewmile: I realized while implementing the redesigned conference page that there is some overlap between `open` and `unclosed` since all open CFPs are also unclosed.  It seemed more useful to me to change the `unclosed` filter to `future`, which still returns unclosed CFPs but they are constrained to CFPs that have not been opened yet.  This leaves no overlap between `open` and `future`.  Does this seem like a good way to approach it?

In the case of the `Api\ConferencesController` if we go with the above, the `unclosed_cfp` case should change to `future_cfp` and the default should probably be `open_cfp`.

mattstauffer: It always hurts my brain a bit to think through all this :)

OK, so you have:

- open CFP (you can apply right now)
- future CFP (you could apply in the future but not right now)

With my original, I thought, you may need different filters depending on what your goal is right now.

You may:
- Want to apply right now (open)
- Want to think through what you want to apply to, whether it's open or opening in the future (unclosed)
- Want to look at what you might apply to in the future (future)

That's my original thoughts... now that you've seen my original thoughts, do you think one of my original categories isn't really that common? I'm fine to change it, just want you to know why I had those categories.

app/Http/Livewire/ConferenceList.php
mattstauffer: Haha, that's awesome!

mattstauffer:
@andrewmile just one note about unclosed cfp; otherwise this is great

app/Http/Controllers/Api/ConferencesController.php
andrewmile: Thanks for laying this out.  I think it comes down to whether or not there is a motivation to view open and future CFP in a single list that outweighs any potential confusion the unclosed filter might have.  I believe this PR replaces the unclosed filter with the future filter so maybe the path of least resistance is to add the unclosed filter back so users will have all three options. Does that sound like a good path forward to you?

PR #491, opened 2023-11-03
Hide past conferences when filtering by open CFP

andrewmile:
This PR modifies the `Open CFP` conference list filter to limit the results to future conferences.  This will prevent past conferences with CFP end dates far into the future from displaying when filtering by open CFP.


mattstauffer:
Good call!

PR #501, opened 2024-04-26
Remove n+1 queries from conference index and conference show pages

andrewmile:
This PR removes n+1 queries with the following solutions:
- The count of open issues is now being loaded on the conferences index and show pages
- Submissions are now loaded on the conferences index page
- Each talk's current revision is now loaded on the conferences index and show pages
- The talk of the current revision is set as a relation on the conference show page

This results in the following query count reduction when testing locally:

| Page | Before | After |
| - | - | - |
| Conference list | 80 | 1 |
| Conference show | 2 | 2 |
| Conference list (auth) | 163 | 6 |
| Conference show | 125 | 7 |

### DIFF
--- a/app/ApiResources/Talk.php
+++ b/app/ApiResources/Talk.php
@@ -27,16 +27,16 @@ class Talk implements Arrayable
     public function attributes()
     {
         return [
-            'title' => $this->talk->currentRevision()->title,
-            'description' => $this->talk->currentRevision()->description,
-            'type' => $this->talk->currentRevision()->type,
-            'length' => $this->talk->currentRevision()->length,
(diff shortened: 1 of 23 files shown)


mattstauffer:
Looks good to me! Great work, and I really appreciate extracting the `currentVersion`; it's somethign I know to do today, thanks to some wonderful database teachers and new Eloquent features, but could only wish for back when I wrote this. (I think I wrote this? And the `current()` is my fault? lol)

I believe this work was co-authored with @omarrida--since he's not marked as a co-author (that's fine, i forget often too), just wanted to credit him in this PR for future knowledge.

